PR #1, opened 2019-08-15
Add search service

emmberk:
Corresponding PR: https://github.com/connexta/ion-store/pull/159


search/src/main/java/com/connexta/search/index/IndexManager.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.connexta.search.index.exceptions.IndexException;
+import java.io.InputStream;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+
+public interface IndexManager {
+
+  void index(
+      @NotBlank String productId, @NotBlank String mediaType, @NotNull InputStream inputStream)
```
ahoffer: The stream is going to get converted into a string in the implementation. Might as well pass the string in directly. The string is human readable (in the debugger) and the stream is not.

search/src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.dao.DataAccessResourceFailureException;
+
+@Slf4j
+public class QueryManagerImpl implements QueryManager {
```
ahoffer: We could make Spring to do the work of the QueryManager.
See https://spring.io/guides/gs/rest-hateoas/
Start reading at the part "So you simply create a plain old java object extending ResourceSupport..."

search/src/main/java/com/connexta/search/query/controllers/QueryController.java
```diff
@@ -0,0 +1,66 @@
...
+      return new ResponseEntity<>(queryManager.find(keyword), HttpStatus.OK);
+    } catch (QueryException e) {
+      log.warn("Unable to search for {}", keyword, e);
+    }
+
+    return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+  }
+
+  @ControllerAdvice
+  private class ConstraintViolationExceptionHandler extends ResponseEntityExceptionHandler {
```
ahoffer: If the @ExceptionHandler is in defined in the controller class, the private class and @ControllerAdvice are unnecessary.

search/src/main/java/com/connexta/search/query/controllers/QueryController.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @ExceptionHandler(value = {ConstraintViolationException.class})
```
ahoffer: If the exception handler only handles a single class of exception, the annotation can be shortened to `@ExceptionHandler(ConstraintViolationException.class)`

search/src/main/java/com/connexta/search/index/IndexManager.java
emmberk: Right now, we just convert it to a string. But in the future, it's going to be json or xml or something else. Wouldn't it make sense to keep it as a InputStream so we're better set up to parse it later?

search/src/main/java/com/connexta/search/query/QueryManagerImpl.java
emmberk: Neat! I will look into that before the mob review.

search/src/main/java/com/connexta/search/query/controllers/QueryController.java
emmberk: I agree, but I just copied this from the mis repo. We can fix it in a future PR.

search/src/main/java/com/connexta/search/query/controllers/QueryController.java
emmberk: https://github.com/connexta/ion-search/pull/1#discussion_r314375500

search/src/main/java/com/connexta/search/query/controllers/QueryController.java
emmberk: `handleExceptionInternal` is defined in ResponseEntityExceptionHandler, so I can't not extend that class, right? In any case, I just copied this from the mis repo, so it can be improved in a future PR to follow what you implemented in https://github.com/connexta/ion-store/pull/148.

build.gradle
```diff
@@ -0,0 +1,106 @@
...
+
+plugins {
+    id "com.diffplug.gradle.spotless" version "3.24.1"
+    id "java"
+    id "maven"
+    id "org.owasp.dependencycheck" version "5.2.1"
+    id "org.sonarqube" version "2.7.1"
+}
+
+allprojects {
```
jhunzik: ✏️ We can move things out of the `allprojects` block. That's only need if you have submodules that you want to apply plugins et. all too.

build.gradle
```diff
@@ -0,0 +1,106 @@
...
+        groovyGradle SpotlessConfig.getGroovy(licenseFile)
+    }
+}
+
+ext.configArgs = [
+        "parallel" : project.hasProperty("parallel"),
+        "skipITests": project.hasProperty("skipITests"),
+]
+
+subprojects {
```
jhunzik: ✏️ The same is true here

search/build.gradle
```diff
@@ -0,0 +1,44 @@
+/*
```
jhunzik: Ah I see why you had the top level `build.gradle` now. Do we need it separated like this now that a repo is essentially one service?

search/src/main/java/com/connexta/search/query/QueryManagerImpl.java
ahoffer: I think we should have Spring HATEOAS for another day. What you've done is fine.

search/src/main/java/com/connexta/search/index/IndexManager.java
ahoffer: I can buy that. :-)

emmberk:
rebased

emmberk:
I've tested:
- `./gradlew clean build`
- `./gradlew dependencyCheckAnalyze --info`
- `docker stack deploy -c docker-compose.yml search-stack` + smoke testing
- `./gradlew deploy` + smoke testing

emmberk:
`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeed for me

emmberk:
Overriding the check for 2 approvers because this has been heroed, is blocking other tasks, and is not about to be released

PR #2, opened 2019-08-19
Fix CircleCI config

emmberk:
@connexta/i2o-cicd


emmberk:
CI will fail until com.connexta.search:index-api-rest-spring-stubs is released

emmberk:
# Do not merge
This is blocked by @shaundmorris.

.circleci/config.yml
```diff
@@ -29,9 +32,9 @@ jobs:
...
-  default:
+  Full build without itests:
     jobs:
-      - build
-      - owasp:
-          requires:
-            - build
\ No newline at end of file
+      - Full build without itests
+  OWASP:
```
ahoffer: This reminds me of something I've been meaning to look into. I've used a single workflow with multiple jobs, but I've never used multiple workflows. Do you know the purpose of multiple workflows? I don't think I found a very good explanation in the Circle CI docs.

emmberk: Workflows can run in parallel

aaronilovici: Also workflows can filter on things like branch. So if you want behavior different for branches (PR vs master) without having to add conditionals to each non-shared step, you can define two different workflows. 

IE workflow A needs to build, test, run owasp, run sonar
    workflow B needs to build, test, tag release, deploy
You won't need to filter each of the non shared steps, you could just have a master workflow and a PR branch workflow. Also can kick them off via cron or other triggers, so you can have a master nightly build and an ad-hoc PR build.

PR #6, opened 2019-08-22
Added Docker image scanning to the CI build

jhunzik:


.circleci/config.yml
```diff
@@ -39,9 +41,54 @@ jobs:
...
     jobs:
       - Full Build
-      - OWASP
\ No newline at end of file
+      - OWASP:
+          requires:
+            - Full Build
+      - Create Docker Image:
+          requires:
+            - OWASP
```
emmberk: ✏️ We don't need to wait until OWASP finishes to scan the image, right? Should this only require Full Build?

.circleci/config.yml
```diff
@@ -39,9 +41,54 @@ jobs:
           name: OWASP
           command: |
             ./gradlew dependencyCheckAnalyze --info
+  Create Docker Image:
```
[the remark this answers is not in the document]
jhunzik: That's right, I just need to figure out where that image is saved so I can scan it with Anchore.

.circleci/config.yml
```diff
@@ -39,9 +48,35 @@ jobs:
...
           command: |
             ./gradlew dependencyCheckAnalyze --info
+  Scan Docker Image:
+    executor: anchore/anchore_engine
+    steps:
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - checkout
+      - attach_workspace:
+          at: /tmp/workspace
```
ahoffer: I've never used workspaces before. Why does the job "Full Build" persist to `workspace`, but the "Scan Docker Image" job attach to `/tmp/workspace`?

PR #14, opened 2019-08-28
Sonarqube

ahoffer:
Circle CI did not pick up previous PR. Opening a new one to attempt to trigger Circle CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,12 +58,12 @@ jobs:
       - run:
           name: Sonar Scan
           command: >
-            ./gradlew sonarqube -x dockerfileZip -x docker -x test
+            ./gradlew sonarqube -x dockerfileZip -x docker -x test --info
             -Dsonar.projectKey=$CIRCLE_PROJECT_REPONAME
             -Dsonar.organization=cx
             -Dsonar.host.url=https://sonarcloud.io
             -Dsonar.java.binaries=**/build/classes
-            -Dsonar.pullrequest.key=$CIRCLE_PR_NUMBER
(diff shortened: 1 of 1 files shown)


ahoffer:
In retrospect, I got a little carried away. Probably best if I just stop changing things.

.circleci/config.yml
```diff
@@ -2,53 +2,80 @@ version: 2.1
...
       - image: circleci/openjdk:11-jdk
     steps:
+      - restore_cache: *restore_cache
       - checkout
       - run:
           name: OWASP
           command: |
-            ./gradlew dependencyCheckAnalyze --info
-  Scan Docker Image:
+            ./gradlew dependencyCheckAnalyze -x test
```
jhunzik: ❓ Find/Replace mistake? The OWASP task doesn't run the test task.

.circleci/config.yml
```diff
@@ -2,53 +2,80 @@ version: 2.1
...
+  Anchore:
```
jhunzik: I'm conflicted, but this matches up better with the Sonar and OWASP

.circleci/config.yml
```diff
@@ -2,53 +2,80 @@ version: 2.1
...
           paths:
             - ion-search.tar
+  Test:
+    machine: *virtual_machine
+    steps:
+      - run: *install_java
+      - run: *configure_java
+      - checkout
+      - restore_cache: *restore_cache
+      - run: ./gradlew check
```
emmberk: ✏️Doesn't the `check` task run the `assemble` task? Looks like it might not make sense to break out the tasks this way.

ahoffer:
Merged a different branch instead of this one. Closing the PR.

PR #19, opened 2019-09-05
Update error handling

ahoffer:
This PR creates a custom error handling class `SearchExceptionHandler` to catch exceptions the normally controller classes would have to catch. The `SearchExceptionHandler` relieves controllers from catching most exception and also from writing to the log.

This PR also creates two exception classes that are not presently used, but whose use is anticipated. They are `MalformedQueryException` and  `IllegalQueryException`. Please read the class comment for each and consider the class names. It is a good time to start thinking about what kind of error conditions can happen during querying. I'm not concerned about the wording of the error messages right now because that is easy to change.


src/main/java/com/connexta/search/common/advice/SearchExceptionHandler.java
```diff
@@ -0,0 +1,53 @@
...
+  @ExceptionHandler(QueryException.class)
+  protected ResponseEntity<Object> handleQueryException(
+      @NotNull final QueryException e, final WebRequest request) {
+    log.warn("Query exception", e);
+    return new ResponseEntity<>(e, e.getStatus());
+  }
+
+  @ExceptionHandler(IndexException.class)
+  protected ResponseEntity<Object> handleIndexException(
+      @NotNull final QueryException e, final WebRequest request) {
```
emmberk: ❗️`IndexException`

src/main/java/com/connexta/search/common/advice/SearchExceptionHandler.java
```diff
@@ -0,0 +1,53 @@
...
+        detailedErrorAttributes.getErrorAttributes(request, false);
+    return new ResponseEntity<>(serializableError, BAD_REQUEST);
+  }
+
+  @ExceptionHandler(QueryException.class)
+  protected ResponseEntity<Object> handleQueryException(
+      @NotNull final QueryException e, final WebRequest request) {
+    log.warn("Query exception", e);
+    return new ResponseEntity<>(e, e.getStatus());
+  }
```
emmberk: ✏️ Move this method to a class in the `com.connexta.search.query` package because it is not common. Same for the method below.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -39,24 +41,25 @@ public void index(
     try {
       idAlreadyExists = crudRepository.existsById(productId);
     } catch (final Exception e) {
-      throw new IndexException("Unable to query index", e);
+      throw new IndexException(INTERNAL_SERVER_ERROR, "Unable to query index", e);
     }
     if (idAlreadyExists) {
-      throw new IndexException("Product already exists. Overriding is not supported");
+      throw new IndexException(
+          INTERNAL_SERVER_ERROR, "Product already exists. Overriding is not supported");
```
emmberk: ✏️ I think this case should return 400. This can probably be handled with a new subclass of `IndexException`, maybe `OverridingIndexException`.

emmberk: ✏️Add a test for this

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -39,24 +41,25 @@ public void index(
...
     }
 
     final Index index;
     try {
       index = new Index(productId, IOUtils.toString(inputStream, StandardCharsets.UTF_8));
     } catch (IOException e) {
-      throw new IndexException("Unable to convert InputStream to String", e);
+      throw new IndexException(INTERNAL_SERVER_ERROR, "Unable to convert InputStream to String", e);
```
emmberk: ✏️ I think this case should return 400.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -39,24 +41,25 @@ public void index(
...
     }
 
     log.info("Attempting to index product id {}", productId);
     try {
       crudRepository.save(index);
     } catch (final Exception e) {
-      throw new IndexException("Unable to save index", e);
+      throw new IndexException(INTERNAL_SERVER_ERROR, "Unable to save index", e);
```
emmberk: ✏️ Create a new subclass of `IndexException`, maybe `DatabaseCommunicationErrorIndexException` for this line and the existsById check.

src/main/java/com/connexta/search/query/controllers/QueryController.java
```diff
@@ -35,25 +27,6 @@
...
-
-    @ExceptionHandler(ConstraintViolationException.class)
-    protected ResponseEntity<Object> handleConstraintViolation(
-        @NotNull final ConstraintViolationException e, @NotNull final WebRequest request) {
-      final String message = e.getMessage();
-      final HttpStatus httpStatus = HttpStatus.BAD_REQUEST;
-      log.warn("Request is invalid: {}. Returning {}.", message, httpStatus, e);
-      return handleExceptionInternal(e, message, new HttpHeaders(), httpStatus, request);
-    }
+    return new ResponseEntity<>(queryManager.find(q), HttpStatus.OK);
```
emmberk: ✏️`return ResponseEntity.ok(queryManager.find(q));`

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -55,10 +55,9 @@ public QueryManagerImpl(
         uri = new URI(endpointUrlRetrieve + id);
       } catch (URISyntaxException e) {
         throw new QueryException(
```
emmberk: ✏️ Should we create a subclass of `QueryException` for when we have communication issues with Solr?

ahoffer: Maybe. Let's see how it goes. We'll know better after spending some time with the services in the cloud.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
ahoffer: Do we have enough evidence to conclude the request was in fact bad? Is there anything about our application that might mishandled the stream along the day?

src/main/java/com/connexta/search/common/advice/SearchExceptionHandler.java
ahoffer: That would mean creating a new class. For one method. If we have plans to break up this repo, I can see the point of it. I like the idea of having all the error handlers for the application in one place.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
ahoffer: I'm going to hold off on new exceptions for now. If IndexException proves too generic, we can add more exception classes.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
ahoffer: Added `testExistingProduct`

src/main/java/com/connexta/search/common/advice/SearchExceptionHandler.java
ahoffer: fixed

PR #21, opened 2019-09-05
Parses the indexed file as JSON

jhunzik:


src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -30,11 +34,14 @@ public IndexManagerImpl(@NotNull final CrudRepository crudRepository) {
...
       @NotBlank final String productId,
       @NotBlank final String mediaType,
-      @NotNull final InputStream inputStream)
+      @NotNull final MultipartFile file)
       throws IndexException {
-    // TODO verify Media Type for CST
     // TODO check that the product exists in S3
 
+    if (!mediaType.equals(MediaType.APPLICATION_JSON_VALUE)) {
+      throw new IndexException("The CST file's media type is not application/json");
```
jhunzik: ✏️ better wording?

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -8,18 +8,22 @@
 
 import com.connexta.search.common.Index;
 import com.connexta.search.index.exceptions.IndexException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.awt.*;
```
jhunzik: I'll fix this

src/main/java/com/connexta/search/index/controllers/IndexController.java
```diff
@@ -40,29 +38,15 @@ public IndexController(@NotNull final IndexManager indexManager) {
     // TODO validate Accept-Version
     // TODO validate productId
 
-    final Long fileSize = file.getSize();
     // TODO validate that fileSize is (0 GB, 10 GB]
 
     final String mediaType = file.getContentType();
     // TODO verify that mediaType is not blank and is a valid Content Type
```
emmberk: ✏️ Remove this TODO

src/main/java/com/connexta/search/index/controllers/IndexController.java
```diff
@@ -40,29 +38,15 @@ public IndexController(@NotNull final IndexManager indexManager) {
...
-          acceptVersion,
-          productId,
-          mediaType,
-          e);
-      return ResponseEntity.badRequest().build();
-    }
-
-    try {
-      indexManager.index(productId, mediaType, inputStream);
+      indexManager.index(productId, mediaType, file);
```
emmberk: Keep MultipartFile only in this class. Don't pass the Spring objects around.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
emmberk: What you have now looks good. Maybe this is better:
```java
"Expected the content type to be " + MediaType.APPLICATION_JSON_VALUE + " but was " + mediaType
```

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -30,11 +33,14 @@ public IndexManagerImpl(@NotNull final CrudRepository crudRepository) {
   public void index(
       @NotBlank final String productId,
       @NotBlank final String mediaType,
-      @NotNull final InputStream inputStream)
+      @NotNull final MultipartFile file)
       throws IndexException {
-    // TODO verify Media Type for CST
     // TODO check that the product exists in S3
 
+    if (!mediaType.equals(MediaType.APPLICATION_JSON_VALUE)) {
```
emmberk: ✏️`StringUtils.equals` to protect against null

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -47,9 +53,10 @@ public void index(
 
     final Index index;
     try {
-      index = new Index(productId, IOUtils.toString(inputStream, StandardCharsets.UTF_8));
+      JsonNode jsonNode = parseJson(file.getInputStream());
+      index = new Index(productId, getElement(jsonNode, EXT_EXTRACTED_TEXT));
```
emmberk: ❓What happens if the json doesn't have a `ext.extracted.text` element? Or the json is empty completely? Or the value for `ext.extracted.text` is an empty string. I think it's worth adding tests for at least these 3 cases.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -59,4 +66,14 @@ public void index(
       throw new IndexException("Unable to save index", e);
     }
   }
+
+  private JsonNode parseJson(InputStream stream) throws IOException {
```
emmberk: ✏️`@NotNull private static`

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -59,4 +66,14 @@ public void index(
...
+    final ObjectMapper objectMapper;
+    objectMapper = new ObjectMapper();
+    return objectMapper.readTree(stream);
+  }
+
+  private String getElement(JsonNode json, String fieldName) {
```
emmberk: ✏️`@NotNull private static` or `@NotBlank` if you add validation for that too

src/main/java/com/connexta/search/index/IndexManagerImpl.java
emmberk: ✏️ Move this Index instantiation out of this try catch. If the Index constructor throws a IOException, we wouldn't want to throw the IndexException in the catch block.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -8,18 +8,21 @@
...
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.io.IOUtils;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.http.MediaType;
+import org.springframework.web.multipart.MultipartFile;
 
 @Slf4j
 public class IndexManagerImpl implements IndexManager {
 
+  public static final String EXT_EXTRACTED_TEXT = "ext.extracted.text";
```
emmberk: ✏️ `private` or use in test classes as well

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -59,4 +66,14 @@ public void index(
...
   }
+
+  private JsonNode parseJson(InputStream stream) throws IOException {
+    final ObjectMapper objectMapper;
+    objectMapper = new ObjectMapper();
+    return objectMapper.readTree(stream);
+  }
+
+  private String getElement(JsonNode json, String fieldName) {
+    return json.get(fieldName).asText();
```
emmberk: ❗️`JsonNode#get(String)` can return `null`

src/main/java/com/connexta/search/index/IndexManagerImpl.java
jhunzik: I added a couple parametrized tests to account for this. One of the Indexing tests that we use a Solr mock has started to fail and I can't figure out why. I disabled the test and added a TODO so that we can revisit it later.

PR #31, opened 2019-09-16
Update the Search service to use OGC CommonQL Queries

jhunzik:


src/main/java/com/connexta/search/query/CommonQl.java
```diff
@@ -39,7 +38,7 @@
    * attributes and throw an exception if an unsupported attribute is detected.
    *
    * @return this for fluent interface
-   * @throws QueryException
+   * @throws IllegalQueryException
```
ahoffer: @jhunzik I think you made the right decision to list the more specific exception (IllegalQueryException). I used QueryException because this method calls get g etFilter which throws a MalformedQueryException. QueryException is the superclass of both.

jhunzik: Ah good catch, I think it's worth listing both exceptions. Right now we only use the super class for 500 type errors.

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -86,6 +90,7 @@ public SolrClientConfiguration testSolrClientConfiguration() {
 
   @BeforeEach
   public void beforeEach() throws IOException, SolrServerException {
+    // TODO shouldn't
```
jhunzik: @emmberk what was this mean?

src/test/java/com/connexta/search/query/CommonQlTest.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.search.query;
+
+import static org.hamcrest.MatcherAssert.*;
```
emmberk: ✏️ Use explicit imports instead of *

src/test/java/com/connexta/search/query/CommonQlTest.java
```diff
@@ -0,0 +1,64 @@
...
+  public static final String QUERY_VALID = "contents LIKE 'bloop'";
+  public static final String QUERY_UNMATCHED_DELIM = "contents LIKE don't";
+  public static final String QUERY_BOOLEAN_UNSUPPORTED_ATTRIBUTE =
+      "contents LIKE 'Winterfell' OR " + INVALID_ATTRIBUTE_NAME + " LIKE 'Kings Landing'";
+
+  @Test
+  public void testValidQuery() {
+    CommonQl object = new CommonQl(QUERY_VALID);
+    object.validate();
+    assertNotNull(object.getFilter());
```
emmberk: ✏️ use hamcrest matchers instead of assertNotNull

src/main/java/com/connexta/search/query/CommonQl.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Class to encapsulate validating a CommonQL query string and building an OCG Filter object from
+ * the string. Create a new instance of this class by passing the CommonQL string into this the
+ * constructor. Please note this class is not thread safe. Each thread should create its own
+ * instance. Also, this class is not meant to be reused with different query strings and that is why
+ * there is not setter for the query string field.
+ */
+@RequiredArgsConstructor
+public class CommonQl {
```
emmberk: ✏️ `CommonQL`

src/main/java/com/connexta/search/query/CommonQl.java
```diff
@@ -0,0 +1,89 @@
...
+  @NotNull private final String queryString;
+  private Filter filter;
+
+  /**
+   * Attempt to validate the query string. Throw an exception is not valid. Check for supported
+   * attributes and throw an exception if an unsupported attribute is detected.
+   *
+   * @return this for fluent interface
+   * @throws IllegalQueryException
+   * @throws MalformedQueryException
```
emmberk: ✏️ Declare these throws in the method signature. Do not include `@throws` in the javadoc if there is no description.

src/main/java/com/connexta/search/query/CommonQl.java
```diff
@@ -0,0 +1,89 @@
...
+   * @returns List of unsupported attributes names found in the query string
+   * @throws MalformedQueryException
+   */
+  @VisibleForTesting
+  Set<String> extractUnsupportedAttributes() {
+    Set<String> properties =
+        Filters.propertyNames(getFilter()).stream()
+            .map(PropertyName::getPropertyName)
+            .collect(Collectors.toSet());
+    properties.removeAll(SolrConfiguration.QUERY_TERMS);
```
emmberk: ✏️ I think you can get the unsupported attributes in one line, e.g.
```java
final Set<String> unsupportedAttributes =
        Filters.propertyNames(filter).stream()
            .map(PropertyName::getPropertyName)
            .filter(p -> !QUERY_TERMS.contains(p))
            .collect(Collectors.toSet());
```
There might be an even easier way to do this in one line.

src/main/java/com/connexta/search/common/configs/SolrClientConfiguration.java
```diff
@@ -1,17 +0,0 @@
-/*
```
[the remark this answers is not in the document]
jhunzik: I thought you deleted that originally. I don't remember removing this.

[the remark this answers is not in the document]
emmberk: 🤦‍♀ I can add it back in.

src/main/java/com/connexta/search/query/CommonQl.java
jhunzik: Declaring them in the method signature results in SonarLint marking it as a minor code smell since it extends a RuntimeException. I'll add a description to each `throws`.

src/main/java/com/connexta/search/query/CommonQl.java
jhunzik: Efficiency! I'll add it in the next commit.

emmberk: I'm actually not sure if it is any more efficient 😆

ahoffer: Emily's code is the pithiest expression I can think of. I don't believe there is an easier way to do it.

src/main/java/com/connexta/search/common/configs/SolrClientConfiguration.java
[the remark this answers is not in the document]
emmberk: Actually, I remember why I made it like this originally. I was thinking that we might have a different kind of solr URL, like one that uses https instead of http. I'll actually simplify the configuration to take just a URL string.

src/main/java/com/connexta/search/query/CommonQl.java
jhunzik: I mean, it's more concise than what I had originally, so I'll take it. 😄

src/test/java/com/connexta/search/query/QueryEndpointTests.java
```diff
@@ -0,0 +1,73 @@
...
+  private static Stream<Arguments> requestsThatThrowErrors() {
+    return Stream.of(
+        Arguments.of(500, new QueryException(HttpStatus.INTERNAL_SERVER_ERROR, "Test")),
+        Arguments.of(400, new IllegalQueryException(Arrays.asList("Test"))),
+        Arguments.of(400, new MalformedQueryException(new RuntimeException())));
+  }
+
+  @Test
+  public void testQueryManagerReturnsList() throws Exception {
+    given(queryManager.find(QUERY_STRING)).willReturn(new ArrayList<>());
```
emmberk: @jhunzik Why do use a new Mockito here? Is this way better? Would it be easy to switch all of the tests to use this?

jhunzik: I'm not sure what you're referring to. This was just just to show the "happy path" when the QueryManager works and when the service should return 200 OK. I figured returning a real empty List would work well for that. If that's what you're talking about, I think I remember having issues with returning a mocked out ArrayList.

emmberk: Sorry for the confusion. Why do you use `org.mockito.BDDMockito` instead of `org.mockito.Mockito`?

|                          |                                                                                                                     |
|--------------------------|---------------------------------------------------------------------------------------------------------------------|
| `org.mockito.BDDMockito` | ```final List queryResults = new ArrayList<>(); given(queryManager.find(QUERY_STRING)).willReturn(queryResults);``` |
| `org.mockito.Mockito`    | ```final List queryResults = new ArrayList<>(); when(queryManager.find(QUERY_STRING)).thenReturn(queryResults);```  |

We use `org.mockito.Mockito` everywhere else in our tests.

jhunzik: Ah that was a mistake. The new Mockito just provides a way to write tests in a behavior driven way, similar to  how we would write it in Spock.

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -45,6 +45,34 @@ public QueryManagerImpl(
     this.dataStore = dataStore;
   }
 
+  /**
```
ahoffer: I had IntelliJ's option to reorder code enabled and this method got moved. I've disabled that in IntelliJ so that it won't happen again.

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -98,31 +130,19 @@ public QueryManagerImpl(
...
-    }
-
-    final Set<String> unsupportedAttributes =
-        Filters.propertyNames(filter).stream()
-            .map(PropertyName::getPropertyName)
-            .filter(attribute -> !QUERY_TERMS.contains(attribute))
-            .collect(Collectors.toSet());
-    if (!unsupportedAttributes.isEmpty()) {
-      throw new IllegalQueryException(unsupportedAttributes);
+  private boolean isKnownError(Exception e) {
```
ahoffer: The the method `isKnownError` is grotesque, but the code to ask Solr if there are any documents in the core is even worse:

```
   CoreAdminRequest req = new CoreAdminRequest();
   req.setAction(CoreAdminParams.CoreAdminAction.STATUS);
   req.setCoreName("searchTerms");
   NamedList<Object> response;
   response = new 
       HttpSolrClient.Builder().withBaseSolrUrl("http://solr:8983/solr/").build().request(req);
   return response.get("status").get("searchTerms").get("index").get("numDocs") > 0
```
And there is a try-catch block for a checked exception that is not shown in this sample code.

emmberk: add a TODO comment to replace this hack

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -135,9 +135,9 @@ private boolean isKnownError(Exception e) {
     // In such a case, return zero results.
     Throwable cause = e.getCause();
     while (cause != null) {
-      if (cause
-          .getMessage()
-          .contains(
+      String msg = cause.getMessage();
+      if (msg != null
+          && msg.contains(
```
emmberk: ✏️ `StringUtils#contains` handles `null`

PR #37, opened 2019-09-25
Added media type as search term

ahoffer:
Index a file.
Then search for it by media type.

~NOTE: Because there is a period in attribute name, it must be enclosed in **double quotes**.~


src/main/java/com/connexta/search/common/configs/SolrConfiguration.java
```diff
@@ -22,8 +22,10 @@
   public static final String SOLR_COLLECTION = "searchTerms";
   public static final String CONTENTS_ATTRIBUTE_NAME = "contents";
   public static final String ID_ATTRIBUTE_NAME = "id";
+  public static final String MEDIA_TYPE_ATTRIBUTE_NAME = "media.type";
```
emmberk: ❗️Update README.md

emmberk: ❗️Add test(s) for querying on this term

emmberk: ✏️ `media_type` or `mediaType` or `mediatype` to eliminate the need to wrap in quotes

ahoffer: I used Susan's CST spreadsheet for the name of the term. If you are comfortable departing from the attribute names she uses, I'll change it to something that doesn't need double quotes.

emmberk: I think we can depart from that list👍

ahoffer: It makes sense to have a test that ensures every supported attribute is indexed properly and can also be queried. I'll add that.

jhunzik: I think it would be a good bet to use the format Solr specifies for field names: `media_type`. It's not enforced, but at least it would be consistent. https://lucene.apache.org/solr/guide/6_6/defining-fields.html

src/main/java/com/connexta/search/query/configs/QueryManagerConfiguration.java
```diff
@@ -50,21 +49,16 @@ public DataStore dataStore(@NotNull final URL solrUrl) throws IOException {
...
-    final List<SolrAttribute> layerAttributes = new ArrayList<>();
 
-    final SolrAttribute contentsSolrAttribute =
-        new SolrAttribute(SolrConfiguration.CONTENTS_ATTRIBUTE_NAME, String.class);
-    contentsSolrAttribute.setEmpty(false);
-    contentsSolrAttribute.setUse(true);
-    layerAttributes.add(contentsSolrAttribute);
+    SolrConfiguration.QUERY_TERMS.stream()
+        .forEach(
+            s -> {
```
jhunzik: ✏️ Change `s` to `term` for clarity.

peterhuffer:
Is there no corresponding solr schema change required?

src/main/java/com/connexta/search/common/configs/SolrConfiguration.java
ahoffer: I'll do that. It's an arbitrary convention, but a a convention is better than no convention. 

I don't believe that Solr will always be the only MIS search engine or database. When additional resources are added to our data layer, we should still be able to stick with the underscore convention, even if it isn't necessarily the convention for those databases/search engines.

src/main/java/com/connexta/search/query/configs/QueryManagerConfiguration.java
ahoffer: In some places we refer to them as "terms" (e.g. searchTerms Solr core). In other places we refer to them as "attributes" (e.g. `CONTENTS_ATTRIBUTE_NAME`). Do "term" and "attribute name" mean the same thing? Or do they mean different things?

jhunzik: Good point, they are "fields" to Solr. Since we're configuring Solr here, it probably makes sense to use the same nomenclature. I don't think OGC CommonQL refers to them as "fields" when creating filters. I think they either refer to them as attributes or properties and that might have caused my confusion.

ahoffer: Most (all?) databases also refer to them as "fields". For this PR, I'll change the local var name to "term" like you suggested. If we want  to standardize our language at point in the future we can. For now, I'll think of "term", "field", "attribute", and "property" as interchangeable.

README.md
```diff
@@ -130,6 +130,7 @@ The following are valid query attributes:
 |---|---|
 | `id` | The unique ID that is associated with each Product |
 | `contents` | A client can perform keyword queries with this attribute. |
+| `media_type` | The media (MIMIE) type of the Product |
```
emmberk: @peterhuffer starting a thread to answer your question https://github.com/connexta/ion-search/pull/37#issuecomment-535076405. We don't have any schema defined currently. The schema is set via the files that Aaron modified in this PR.

src/main/java/com/connexta/search/index/Index.java
```diff
@@ -26,4 +26,7 @@
 
   @Indexed(name = SolrConfiguration.CONTENTS_ATTRIBUTE_NAME, type = "string")
   private String contents;
+
+  @Indexed(name = SolrConfiguration.MEDIA_TYPE_ATTRIBUTE_NAME, type = "string")
```
ahoffer: @peterhuffer The Sprint Data library handles the Solr schema. The @Indexed annotation is part of that library.

src/main/java/com/connexta/search/index/IndexManagerImpl.java
```diff
@@ -67,7 +67,7 @@ public void index(
 
     log.info("Attempting to index product id {}", productId);
     try {
-      crudRepository.save(new Index(productId, contents));
+      crudRepository.save(new Index(productId, contents, mediaType));
```
ahoffer: @connexta/i2o-cdr Passing every attribute into Index constructor won't look good after a bunch of terms are added. Add some point another strategy is needed, like setters or a builder.

src/main/java/com/connexta/search/query/configs/QueryManagerConfiguration.java
```diff
@@ -43,28 +42,21 @@ public DataStore dataStore(@NotNull final URL solrUrl) throws IOException {
...
+              SolrAttribute a = new SolrAttribute(s, String.class);
+              a.setEmpty(false);
+              a.setUse(true);
+              solrLayerConfiguration.getAttributes().add(a);
+            });
```
ahoffer: @peterhuffer On the query side, we convert a CommonQL string to a OGC filter. An additional library is used to bridge to Solr. It accepts an OGC filter and returns a list of OGC features, which are the search results. However, we have to configure that library with every Solr field that is part of the feature type. That is done here.

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+public class SearchTermITests {
+
+  private static final int SOLR_PORT = 8983;
+
+  @Container
+  public static final GenericContainer solrContainer =
+      new GenericContainer("solr:8.1.1")
+          .withCommand("solr-create -c " + SOLR_COLLECTION)
+          .withExposedPorts(SOLR_PORT)
+          // TODO: would (new URL("solr", SOLR_COLLECTION, "admin", "ping").toString() work?
```
emmberk: I don't think that would work because `new URL("solr", SOLR_COLLECTION, "admin", "ping")` is not a valid URL.

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+          .waitingFor(Wait.forHttp("/solr/" + SOLR_COLLECTION + "/admin/ping"));
+
+  @Inject private QueryManager queryManager;
+  @Inject private IndexManager indexManager;
+  private Map<String, String> sampleDocument = createSampleIndexData();
+
+  public SearchTermITests() throws IOException {}
+
+  @NotNull
+  private Map<String, String> createSampleIndexData() throws IOException {
```
emmberk: ✏️ Inline this because it's only used once
```java
private static final Map<String, String> sampleDocument =
    Map.of(
        ID_ATTRIBUTE_NAME,
        UUID.randomUUID().toString(),
        CONTENTS_ATTRIBUTE_NAME,
        "Winterfell",
        MEDIA_TYPE_ATTRIBUTE_NAME,
        "application/json");
```

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+    return Map.of(
+        ID_ATTRIBUTE_NAME,
+        UUID.randomUUID().toString(),
+        CONTENTS_ATTRIBUTE_NAME,
+        "Winterfell",
+        MEDIA_TYPE_ATTRIBUTE_NAME,
+        "application/json");
+  }
+
+  private String getQuery() {
```
emmberk: ✏️ Inline this because it's only used once

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+  public static final GenericContainer solrContainer =
+      new GenericContainer("solr:8.1.1")
+          .withCommand("solr-create -c " + SOLR_COLLECTION)
+          .withExposedPorts(SOLR_PORT)
+          // TODO: would (new URL("solr", SOLR_COLLECTION, "admin", "ping").toString() work?
+          .waitingFor(Wait.forHttp("/solr/" + SOLR_COLLECTION + "/admin/ping"));
+
+  @Inject private QueryManager queryManager;
+  @Inject private IndexManager indexManager;
+  private Map<String, String> sampleDocument = createSampleIndexData();
```
emmberk: ✏️ `private static final`

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+
+  public SearchTermITests() throws IOException {}
```
emmberk: ✏️ Remove this constructor

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+        .map(term -> String.format("%s = '%s'", term, sampleDocument.get(term)))
+        .collect(Collectors.joining(" AND "));
+  }
+
+  InputStream getMetadata() throws IOException {
+    return IOUtils.toInputStream("{ \"ext.extracted.text\" : \"Winterfell\" }", "UTF-8");
+  }
+
+  @Test
+  void verifySampleDataComplete() {
```
emmberk: ❓ What is this test verifying? It looks like you're only checking that you instantiated `sampleDocument` correctly. Does it verify anything in the app? If not, I don't think this test provides value.

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+    assertThat(
+        "Sample document is not complete", QUERY_TERMS.equals(sampleDocument.keySet()), is(true));
+  }
+
+  @Test
+  public void test() throws IOException {
```
emmberk: ✏️ Rename this to `testAllDefinedQueryTermsAreSupported`

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  private String getQuery() {
+    return QUERY_TERMS.stream()
+        .map(term -> String.format("%s = '%s'", term, sampleDocument.get(term)))
+        .collect(Collectors.joining(" AND "));
+  }
+
+  InputStream getMetadata() throws IOException {
+    return IOUtils.toInputStream("{ \"ext.extracted.text\" : \"Winterfell\" }", "UTF-8");
```
emmberk: ✏️ `IndexManagerImpl.EXT_EXTRACTED_TEXT`

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+  @Test
+  void verifySampleDataComplete() {
+    assertThat(
+        "Sample document is not complete", QUERY_TERMS.equals(sampleDocument.keySet()), is(true));
+  }
+
+  @Test
+  public void test() throws IOException {
+    ObjectMapper mapper = new ObjectMapper();
+    Index document = mapper.convertValue(sampleDocument, Index.class);
```
emmberk: ✏️ Why do you need to map this? `sampleDocument` is only used to hold pairs of terms and values, right? Would it be simpler to pass your tests values directly to indexManager#index?

src/test/java/com/connexta/search/common/SearchTermITests.java
emmberk: ✏️ Can this test be moved to `SearchITests` instead? That way, we don't need to start up the Solr container another time. Or maybe move the testQueryXxx methods from SearchITests to this class.

src/main/java/com/connexta/search/index/Index.java
```diff
@@ -6,8 +6,14 @@
...
 package com.connexta.search.index;
 
+import static com.connexta.search.common.configs.SolrConfiguration.CONTENTS_ATTRIBUTE_NAME;
+import static com.connexta.search.common.configs.SolrConfiguration.ID_ATTRIBUTE_NAME;
+import static com.connexta.search.common.configs.SolrConfiguration.MEDIA_TYPE_ATTRIBUTE_NAME;
+
 import com.connexta.search.common.configs.SolrConfiguration;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.AllArgsConstructor;
+import lombok.Builder;
```
mcalcote: I got a spotless failure on this file for some reason 🤔 ... Not sure if it an actual issue.

src/test/java/com/connexta/search/common/SearchTermITests.java
mcalcote: `java.lang.IllegalArgumentException: Unrecognized field "media_type" (class com.connexta.search.index.Index), not marked as ignorable (3 known properties: "contents", "id", "mediaType"])
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: com.connexta.search.index.Index["media_type"])`

src/test/java/com/connexta/search/common/SearchTermITests.java
```diff
@@ -0,0 +1,120 @@
...
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(webEnvironment = WebEnvironment.DEFINED_PORT)
+@Testcontainers
+@DirtiesContext
+public class SearchTermITests {
+
+  private static final int SOLR_PORT = 8983;
+
+  @Container
+  public static final GenericContainer solrContainer =
```
peterhuffer: ✏️ We can optionally use an embedded solr server.

src/test/java/com/connexta/search/common/SearchTermITests.java
emmberk: I got this failure too.

README.md
peterhuffer: ✏️ `MIME`

src/test/java/com/connexta/search/common/SearchTermITests.java
emmberk: I think we should hold off on that for now. We have a ticket to add that improvement in our backlog.

src/main/java/com/connexta/search/index/Index.java
peterhuffer: I just started the stack and hit the solr admin UI and looked at the `searchTerms` collections schema. These annotations do not create the fields. It looks like we are using the default solr schema, which is relying on dynamic fields. At some point we probably don't want dynamic fields and want to define our own unless we expect to index and search on arbitrary fields.

peterhuffer: The default solr schema also has some comments about removing some sections for performance reasons if they are not needed.

emmberk: Yeah. This is a known issue. There are currently issues when we query before anything is stored.

README.md
ahoffer: LOL. Thanks Peter.

src/main/java/com/connexta/search/index/Index.java
ahoffer: That was supposed to have been gone. I hope I didn't fudge up the commits.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: Thanks. I remember that one. And I see how it got reintroduced when I did some cleanup.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: What I'm about to write is off-topic. Also, this isn't the right the place for the discussion, but putting our integration tests inside of the code base itself has always given me heart burn -- even in early DDF days. I like unit tests to accompany the source code, but I wish it stopped there. I would rather spin up containers, services, or servers and test them with an external tool. </rant>

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: I like the test data isolated from the test case that is using it.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: I like it encapsulated in its own method.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: Deleted the variable when I moved the test over into the SearchITest class.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: Good point. The query will fail if any of the approved query terms from missing from the test data.

src/test/java/com/connexta/search/common/SearchTermITests.java
peterhuffer: I do not think it is bad to have integration tests live next to our code. Integration tests in the new sense is not setting up a full instance like DDF integration tests. Plus we just have to make sure ours are not slow and flaky unlike DDF!

ahoffer: @peterhuffer I think the world agrees with you. Otherwise things like Pax Exam and Spring Boot Test  would not exist, right? 
My contrarian opinion comes from having been tripped up on things like subtle differences between RestTemplate and TestRestTemplate. In another case, we also spend several days figuring why an itest would not work. Eventually, Josh stuck a `@DirtiesContext` annotation on the test and everything worked perfectly. 
I feel like Pax Exam and Spring Boot Test lure you in with the promise of easier integration tests, but the little gotchas end up costing you more time than you expected. Investing in a purpose-built tool has its own drawbacks, but it is common thing to do in the industry.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: `EXT_EXTRACTED_TEXT` is a private static field of `IndexManagerImpl. The test class would not be able to access it. Also, `EXT_EXTRACTED_TEXT` is only used once. If the letters are changed to lower case and the underscores with periods, EXT_EXTRACTED_TEXT` becomes `ext.extracted.text`. That makes `EXT_EXTRACTED_TEXT` a good candidate for in-lining. 

Longer-term, we create a class that models the metadata we expect to get from transform. Then the JSON library will serialize and de-serialize it for us-- no more passing around input streams.

src/test/java/com/connexta/search/common/SearchTermITests.java
ahoffer: Good catch. That was left-over from some earlier code. I assumed an IndexManager's public method would accept an instance of Index. It did not. 

After looking through the IndexManagerImpl class it has two separate kinds of responsibilities:
1. Construct  an Index object from other objects (strings and a stream).
2. Talk to a Repository to determine if the item is allowed to be stored, and then storing it.

Responsibility is the builder pattern. It does not belong inside the manager class. The Index should be built outside the manager and passed into it.

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -70,25 +81,39 @@
...
-          "http",
-          solrContainer.getContainerIpAddress(),
-          solrContainer.getMappedPort(SOLR_PORT),
-          "/solr");
-    }
-  }
-
+  private static final String INDEX_ENDPOINT_BASE_URL = "/index/";
+  @Inject private QueryManager queryManager;
+  @Inject private IndexManager indexManager;
```
emmberk: ✏️ Doesn't need to be addressed in this PR, but I don't think these beans should not be injected here. I think it's easier to read, maintain, and debug if all the tests in a test class consistently use only external endpoints or only beans.

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -358,27 +383,57 @@ public void testQueryWhenSolrIsEmpty() throws Exception {
...
-          public String getFilename() {
-            // The extension of this filename is used to get the ContentType of the file.
-            return "ignored.json";
-          }
-        });
-    final HttpHeaders httpHeaders = new HttpHeaders();
-    httpHeaders.set("Accept-Version", "0.2.0");
-    return new HttpEntity<>(requestBody, httpHeaders);
+  @NotNull
+  private Map<String, String> getSampleDatatHavingAllAttributes() {
```
emmberk: ✏️`static`
✏️Does this need to be a method? Would it be simpler to use of a `private static final` field because it's not dynamic?

(same comments for `getQueryForAllAttributes`)

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -358,27 +383,57 @@ public void testQueryWhenSolrIsEmpty() throws Exception {
...
 
-          @Override
-          public long contentLength() throws IOException {
-            return metadataInputStream.available();
-          }
+    // Assert valid preconditions
+    assertThat(
+        "Sample data must include all query attributes",
+        unmodifiableSet(doc.keySet()),
+        equalTo(QUERY_TERMS));
```
emmberk: ✏️I think this verification of your test data is unnecessary

src/test/java/com/connexta/search/common/SearchITests.java
ahoffer: Good suggestion. They look better as vars.

PR #38, opened 2019-09-26
Allow compiling with java11+

peterhuffer:
Remove the java version check. Compiling to java 11 should be supported by java 11,12,13,14 based on https://openjdk.java.net/jeps/182

This check is already also done by java and will print out a message like `Could not target platform: 'Java SE 11' using tool chain: 'JDK 8 (1.8)'.` if the installed JDK cannot target the target version.

@emmberk @ahoffer @jhunzik

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,8 +5,6 @@
  * https://www.gnu.org/licenses/lgpl-3.0.html
  */
 /* Build Script */
-import org.gradle.internal.jvm.Jvm
-
 buildscript {
     repositories {
         maven { url "https://plugins.gradle.org/m2/" }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -70,8 +70,8 @@ dependencies {
 sourceCompatibility = 1.11
 targetCompatibility = 1.11
 
-if (sourceCompatibility != Jvm.current().javaVersion) {
```
peterhuffer: We could also just remove this check and let java handle this. It would print something like
`Could not target platform: 'Java SE 11' using tool chain: 'JDK 8 (1.8)'.`

emmberk: We initially did not have this check, but a few devs reached out to us to ask why they couldn't build, which means that the default error message and a note in the README wasn't sufficient.

build.gradle
```diff
@@ -70,8 +70,8 @@ dependencies {
...
-    throw new Exception("You need Java ${sourceCompatibility} to build and run ${project.name}.\n" +
+if (Jvm.current().javaVersion < sourceCompatibility) {
```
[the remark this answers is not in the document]
peterhuffer: Yes because later java versions will compile to the target version (which will always be 11 in our case)

emmberk: ✏️ Update README

ahoffer:
1. All Sonar Cloud jobs are failing temporarily disabling merge checks.
2. Peter asked to preserve the commits. Temporarily enabling "rebase and merge"
3. Merging.

PR #40, opened 2019-09-30
Upgrade gradle wrapper to use 5.6.2

peterhuffer:
Also fixes gradlew generated by gradle v5.4.1 not working on windows.
Without this change windows gets the following error when using the gradle wrapper
```
Error: Could not find or load main class
Caused by: java.lang.ClassNotFoundException:
```
This PR is the output of running `gradle wrapper`

@mcalcote @emmberk @jhunzik 

Hero:
On windows run `.\gradlew build` and verify it builds.

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -5,7 +5,7 @@
 @rem you may not use this file except in compliance with the License.
 @rem You may obtain a copy of the License at
 @rem
-@rem      http://www.apache.org/licenses/LICENSE-2.0
+@rem      https://www.apache.org/licenses/LICENSE-2.0
 @rem
 @rem Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


gradlew.bat
```diff
@@ -30,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=""
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
mcalcote: is this right?

peterhuffer: Yes

mcalcote: 🏰

PR #54, opened 2019-10-10
Extract indexApiVersion

emmberk:


src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -429,4 +409,28 @@ public URL solrUrl() throws MalformedURLException {
...
+          }
+
+          @Override
+          public String getFilename() {
+            // The extension of this filename is used to get the ContentType of the file.
+            return "ignored.json";
+          }
+        });
+    final HttpHeaders httpHeaders = new HttpHeaders();
+    httpHeaders.set(IndexController.ACCEPT_VERSION_HEADER_NAME, indexApiVersion);
```
mcalcote: ✏️ Since this is the only use of IndexController just static import ACCEPT_VERSION_HEADER_NAME

src/test/java/com/connexta/search/index/IndexTests.java
```diff
@@ -18,11 +18,11 @@
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.multipart;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
+import com.connexta.search.index.controllers.IndexController;
```
mcalcote: ✏️ Same comment here -> Since this is the only ACCEPT_VERSION_HEADER_NAME is used from IndexController just static import

src/test/java/com/connexta/search/index/IndexTests.java
```diff
@@ -75,24 +75,115 @@ public void after() {
...
         .perform(
-            requestBuilder
+            multipart("/index/1234")
+                .file(
+                    new MockMultipartFile(
+                        "file",
+                        "this originalFilename is ignored",
+                        "application/json",
+                        IOUtils.toInputStream(
+                            "{\"ext.extracted.text\" : \"All the color had been leached from Winterfell until only grey and white remained\"}",
```
mcalcote: “Hodor," said Hodor.

src/test/java/com/connexta/search/common/SearchITests.java
emmberk: The reason I chose to keep it this way is because we probably will add an Accept-Version to the QueryApi, and I wanted to be explicit that this field should only be used for the IndexApi. An alternative would be naming the field INDEX_API_ACCEPT_VERSION_HEADER_NAME.

src/test/java/com/connexta/search/index/IndexTests.java
emmberk: See https://github.com/connexta/ion-search/pull/54#discussion_r335072853.

emmberk:
I heroed this locally.

```
Expected Accept-Version to be "0.2.0" but was "0.5.0". Only "0.2.0" is currently supported.
```

PR #57, opened 2019-10-22
Create a local deployment script and update exception handler.

jhunzik:
This PR should be heroed to make sure that the error responses are consistent with the other services.

Situations to be tested:

- invalid CommonQL string
- no `q` query parameter
- Querying when solr is not connected
- Indexing when solr is not connected
- Indexing without metadata 
- any other use cases


src/main/java/com/connexta/search/common/advice/CustomExceptionHandler.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @ExceptionHandler(ConstraintViolationException.class)
+  protected ResponseEntity<Object> handleConstraintViolation(
+      @NotNull final ConstraintViolationException e, @NotNull final WebRequest request) {
+    final HttpStatus status = HttpStatus.BAD_REQUEST;
+    log.warn("Request is invalid. Returning {}.", status, e);
+    final Map<String, Object> errorAttributes =
+        detailedErrorAttributes.getErrorAttributes(request, false);
+    errorAttributes.put("status", status.value());
+    errorAttributes.put("error", status.getReasonPhrase());
```
peterhuffer: ❓ What will this `status.getReasonPhrase()` be since we just did a straight `final HttpStatus status = HttpStatus.BAD_REQUEST` right above?

jhunzik: The value refers to the actual status code and the reason phrase would be `BAD REQUEST`

peterhuffer: Should we be cycling through the Constraint violations on the exception and grabbing those messages instead?

jhunzik: Yeah, there's definitely room for improvement with what we send back in the response. This change was to just make sure we actually send back a formatted response.

peterhuffer:
Hero success I think! ✔️ 

For invalid request `http://localhost:9042/search?q=thisisn'tvalid`
I got HTTP 400 `Invalid CommonQL query string`

For no parameter request `http://localhost:9042/search`
I got 400 `Required String parameter 'q' is not present`

With docker service shutdown and request `http://localhost:9042/q=contents%20LIKE%20'Winterfell'`
I got a HTTP 500 - Unable to search for contents LIKE 'Winterfell'

For index without solr running
I got HTTP 500 - Unable to query index

For index without metadata
Got a message HTTP 415 - Content type '' not supported

PR #58, opened 2019-10-24
Fix solr volume not persisting

peterhuffer:
Fixes solr not persisting volumes between container restarts.

Hero:
- Start the search-stack.
- Using postman or the solr admin UI, create a configuration
- Search solr through the admin ui to make sure the config is saved
- Stop the stack
- Start the stack
- Search solr through the admin ui to make sure the previous config is still there

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
     ports:
       - "9983:8983"
     volumes:
-      - data:/opt/solr/server/solr/mycores
+      - data:/var/solr
     entrypoint:
       - docker-entrypoint.sh
       - solr-precreate


docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
     ports:
       - "9983:8983"
     volumes:
-      - data:/opt/solr/server/solr/mycores
```
peterhuffer: TBH, I have no idea why this fixed it, but I was able to start solr, create a config, stop solr, start it, then search for the config. I thought the issue was with the entry point originally.

peterhuffer: I didn't have much luck finding out what happens differently with these 2 paths either when observing the service's logs.

figliold: I have seen strange behavior (permission issues, etc.) if one attempts to mount a volume in a directory that already exists?  Is that the case here?

jhunzik: Should we be adding that deployment constraint that Chris mentioned to ensure we're accessing the same volume on the same host every time?

peterhuffer: Yeah I believe that `/var/solr` is the default directory in the solr image.

PR #61, opened 2019-10-28
Fix remote deploy solr persistence

peterhuffer:
Fixes solr data not persisting between service restarts. (not sure why this fixes the issue but it does)

@emmberk @ahoffer @jhunzik 

To hero:
Deploy remotely
Add an entry to solr
Restart the stack
Query solr and verify the previous entry is there

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
     ports:
       - "9983:8983"
     volumes:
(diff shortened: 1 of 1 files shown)


peterhuffer:
Overriding CI. Full build passed for CircleCI is not kicking off owasp build.

PR #62, opened 2019-10-28
Initial Threat Model

mcalcote:
Open in [Threat Dragon](https://docs.threatdragon.org/#downloads) to Review
image


ThreatModels/Search.json
```diff
@@ -1 +1,590 @@
...
+                  "y": 165
+                }
+              ],
+              "id": "3b2a0fd7-d7cc-4443-8a10-9912a6fa934a",
+              "labels": [
+                {
+                  "position": 0.5,
+                  "attrs": {
+                    "text": {
+                      "text": "Fetch File",
```
ahoffer: Sounds like we will have a group review of threat models. I'll use GitHub to jot down questions I'd like to ask when that happens.
What is the "Fetch file" message?

mcalcote: It is a GET to the StoreController with the product ID... This isn't implemented yet as far as I can tell, but @emmberk said it needed to be included.

mcalcote: Actually that would be the `Retrieve Request` portion, but the fetch file message would be a 200 Response with the content's media type in the header and the retrieve response stream in the body.

emmberk:
index:
index

query:
query

PR #63, opened 2019-10-30
Return links to IRM results from the query endpoint

figliold:
Return links to IRM results from the query endpoint

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ The following are valid query attributes:
 |---|---|
 | `id` | The unique ID that is associated with each dataset. |
 | `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the file. |
+| `media_type` | The media (MIME) type of the IRM. |
 
 ### Inspecting
 The service is deployed with (Springfox) **Swagger UI**.
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ For more information about spotless checks see
     Example configs/search_config.yml:
     ```yaml
     endpointUrl:
-      productRetrieve: http://localhost:9041/mis/product/
+      datasetRetrieve: http://localhost:9041/dataset
     ```
 
 2. A Docker network named `cdr` is needed to run via docker-compose.
@@ -120,7 +120,7 @@ After configuring the build system:
 
 ## Using
 ### Querying
-The Search service can be queried for Products. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
+The Search service can be queried for IRM. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
 The Search service supports the OGC Catalogue Common Query Language (OGC CommonQL).
 See [Annex B - BNF Definition of OGC CommonQL](http://docs.opengeospatial.org/is/12-168r6/12-168r6.html#62) for the definition of the grammar.
 
@@ -128,9 +128,9 @@ The following are valid query attributes:
 
 | Attribute | Description  |
 |---|---|
-| `id` | The unique ID that is associated with each Product |
+| `id` | The unique ID that is associated with each IRM |
 | `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the Product |
+| `media_type` | The media (MIME) type of the IRM |
 
 ### Inspecting
 The service is deployed with (Springfox) **Swagger UI**.
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ After configuring the build system:
 
 ## Using
 ### Querying
-The Search service can be queried for IRM. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
+The Search service can be queried for datasets. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters.
 The Search service supports the OGC Catalogue Common Query Language (OGC CommonQL).
 See [Annex B - BNF Definition of OGC CommonQL](http://docs.opengeospatial.org/is/12-168r6/12-168r6.html#62) for the definition of the grammar.
 
@@ -128,9 +128,9 @@ The following are valid query attributes:
 
 | Attribute | Description  |
 |---|---|
-| `id` | The unique ID that is associated with each IRM |
+| `id` | The unique ID that is associated with each dataset. |
 | `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the IRM |
+| `media_type` | The media (MIME) type of the file. |
 
 ### Inspecting
 The service is deployed with (Springfox) **Swagger UI**.
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,6 @@
 buildscript {
     ext {
         indexApiVersion = "0.2.0"
-        queryApiVersion = "0.3.0"
         junitJupiterVersion = "5.5.2"
         lombokVersion = "1.18.10"
         springBootVersion = "2.2.0.RELEASE"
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -120,17 +120,17 @@ After configuring the build system:
 
 ## Using
 ### Querying
-The Search service can be queried for Products. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
+The Search service can be queried for IRM. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
```
ahoffer: Food for thought. 
From a "RESTful" perspective, the Search service is querying for a resource. In this case the resource is "dataset". 
The most RESTful thing to do is to return a URI to the dataset itself. 

In a RESTful paradigm, IRM is an XML document that is one "representation" of a dataset. To get an IRM document, the client would send a GET to the dataset's URI with an accept header. The message would look like:

> GET /dataset
> HOST store.com
> ACCEPT: irm/xml

The media type is `irm/xml`. I believe that applications should come up with media types meaningful their applications. It makes a more complete model of the domain. 

Another alternative would be to use a query param,  but it is "RESTful"

`http://host/dataset?representation=irm`

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -104,24 +105,26 @@ static Filter getFilter(final String queryString)
...
    */
-  private List<URI> getProductUris(final List<String> matchingIds) {
+  private List<URI> getIrmUris(final List<String> matchingIds) {
     return matchingIds.stream()
         .map(
             id -> {
               try {
-                return new URI(productRetrieveEndpoint + id);
+                final String irmRetrieveEndpoint =
+                    String.format("%s/%s/%s", datasetRetrieveEndpoint, id, IRM_PATH_SEGMENT);
```
ahoffer: It is safer to use something like the [UriComponentBuilder](https://www.baeldung.com/spring-uricomponentsbuilder) to construct URIs than to use template strings (or string concatenation). Classes like UriComponentBuilder and Java's URI understand the domain of URIs. They help ensure that only valid URIs are constructed, and that characters and strings are properly escaped.

README.md
emmberk: We cannot modify the API right now because it would break the thread. We have a ticket in our next sprint to re-write the Store API. I will copy your comment there.

README.md
```diff
@@ -61,7 +61,7 @@ For more information about spotless checks see
     Example configs/search_config.yml:
     ```yaml
     endpointUrl:
-      productRetrieve: http://localhost:9041/mis/product/
+      datasetRetrieve: http://localhost:9041/dataset
```
emmberk: ❗️Update the rest of the repo to use the right terms. E.g.
- `product` should never be used anywhere in the repo
- `datasetId` instead of `id`

build.gradle
```diff
@@ -8,6 +8,7 @@
 buildscript {
     ext {
         indexApiVersion = "0.2.0"
+        queryApiVersion = "0.3.0"
```
emmberk: ✏️ Right now, we only extract version properties that are used in more than one place. Since this is only used once, it can be inlined.

README.md
```diff
@@ -120,17 +120,17 @@ After configuring the build system:
...
+The Search service can be queried for IRM. See [Query Service API](https://github.com/connexta/ion-query-api) for information about paths and parameters
 The Search service supports the OGC Catalogue Common Query Language (OGC CommonQL).
 See [Annex B - BNF Definition of OGC CommonQL](http://docs.opengeospatial.org/is/12-168r6/12-168r6.html#62) for the definition of the grammar.
 
 The following are valid query attributes:
 
 | Attribute | Description  |
 |---|---|
-| `id` | The unique ID that is associated with each Product |
+| `id` | The unique ID that is associated with each IRM |
```
emmberk: ❗️`dataset`

README.md
```diff
@@ -120,17 +120,17 @@ After configuring the build system:
...
 | `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the Product |
+| `media_type` | The media (MIME) type of the IRM |
```
emmberk: ❗️`file`

README.md
emmberk: ✏️ Missing a period at the end of this sentence.

emmberk: ✏️ `datasets` instead of `IRM`

src/main/java/com/connexta/search/query/QueryManagerImpl.java
```diff
@@ -36,12 +36,13 @@
...
+  @NotBlank private final String datasetRetrieveEndpoint;
   @NotNull private final DataStore dataStore;
 
   public QueryManagerImpl(
-      @NotNull final DataStore dataStore, @NotBlank final String productRetrieveEndpoint) {
-    this.productRetrieveEndpoint = productRetrieveEndpoint;
+      @NotNull final DataStore dataStore, @NotBlank final String datasetRetrieveEndpoint) {
+    this.datasetRetrieveEndpoint = datasetRetrieveEndpoint;
     this.dataStore = dataStore;
   }
```
emmberk: ✏️ `@AllArgsConstructor`

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -206,19 +214,22 @@ public void testStoreMetadataCstWhenSolrIsNotEmpty() throws Exception {
...
     assertThat(
         (List<String>) restTemplate.getForObject(queryUriBuilder.build(), List.class),
-        hasItem(equalTo(productLocation)));
+        hasItem(equalTo(irmLocation)));
   }
 
   @Test
-  @Disabled("TODO check that the product exists before storing cst")
-  public void testStoreMetadataProductIdNotFound() {}
+  @Disabled("TODO check that the IRM exists before storing cst")
```
emmberk: ❗️`@Disabled("TODO check that the dataset exists before indexing")`

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -206,19 +214,22 @@ public void testStoreMetadataCstWhenSolrIsNotEmpty() throws Exception {
...
+  public void testStoreMetadataIrmIdNotFound() {}
```
emmberk: ❗️`datasetId`

src/test/java/com/connexta/search/common/SearchITests.java
```diff
@@ -206,19 +214,22 @@ public void testStoreMetadataCstWhenSolrIsNotEmpty() throws Exception {
...
 
   @Test
-  public void testStoreWhenProductHasAlreadyBeenIndexed() throws Exception {
-    // given index a product
+  public void testStoreWhenIrmHasAlreadyBeenIndexed() throws Exception {
```
emmberk: ❗️ `testStoreWhenDatasetHasAlreadyBeenIndexed`

README.md
emmberk: Actually, this _is_ currently the mediatype of the file attachment, which is the IRM. I will change this back. 🤦‍♀

build.gradle
```diff
@@ -41,9 +41,10 @@ dependencies {
     annotationProcessor "org.projectlombok:lombok:${lombokVersion}"
     compileOnly "org.projectlombok:lombok:${lombokVersion}"
     implementation "com.connexta.search:index-api-rest-spring-stubs:${indexApiVersion}"
-    implementation "com.connexta.search:query-api-rest-spring-stubs:0.2.0"
+    implementation "com.connexta.search:query-api-rest-spring-stubs:0.3.0"
     implementation "io.springfox:springfox-swagger2:${springfoxSwaggerVersion}"
     implementation "io.springfox:springfox-swagger-ui:${springfoxSwaggerVersion}"
+    implementation "org.springframework:spring-web:5.2.0.RELEASE"
```
emmberk: ❓Why was this added?

emmberk: I was able to full build without this. If it is needed, it should be caught in the hero.

emmberk:
@ahoffer is going to hero this.

ahoffer:
I have heroed this locally.
1. I used the index endpoint to create an entry.
2. I used the query endpoint to search for the entry.
3. The endpoint returns a JSON array of of (1) URI

Here is the screen shot from Postman
image

PR #69, opened 2019-11-04
Improve input validation

emmberk:
This PR also
- renames `Manager`s to `Service`s to follow Spring conventions
- fixes local deployment


emmberk:
I heroed this PR locally by verifying the error json when
- missing file param
- file is missing
- Accept-version is missing
- Accept-Version is not 0.2.0

src/test/java/com/connexta/search/query/QueryControllerTest.java
```diff
@@ -31,8 +31,8 @@
 import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
 
 /**
- * This class uses WebMvc to test the controller class. The purpose is to test the controller. The
- * query manager service that supports the controller is mocked.
+ * This class uses {@code @WebMvcTest} to test {@link QueryController}. The query manager service
```
jhunzik: ✏️ `query manager`

src/test/java/com/connexta/search/query/QueryServiceTest.java
```diff
@@ -42,9 +42,10 @@
 import org.springframework.web.util.UriComponentsBuilder;
 
 /**
- * This class uses Sprint Boot Test to fully start the container. The primary purpose is to test the
- * Query Manager Service, but the test requests go through the controller to reach the query
- * manager. The datastore that supports the query query is mocked.
+ * This class uses {@link SpringBootTest} to fully start the container. The primary purpose is to
+ * test the {@link QueryService}, but the test requests go through the {@link
+ * com.connexta.search.query.controllers.QueryController} to reach the query manager. The {@link
```
jhunzik: ✏️ `query service`

src/main/java/com/connexta/search/index/controllers/MultipartFileValidator.java
```diff
@@ -16,47 +16,45 @@
...
-    this.file = file;
-  }
+  private MultipartFileValidator() {}
 
-  public void validate() {
-    validateSize();
-    validateContentType();
-    validateFilename();
-    validateInputStream();
+  public static MultipartFile validate(final MultipartFile file) {
```
ahoffer: You might want to make this an instance method so the validator can be injected as a bean.

src/test/java/com/connexta/search/index/IndexTests.java
```diff
@@ -91,6 +92,33 @@ void testMissingFile() throws Exception {
         .andExpect(status().isBadRequest());
   }
 
+  @ParameterizedTest(name = "400 Bad Request if ID is {0}")
+  @NullSource
+  @ValueSource(
+      strings = {"   ", "1234567890123456789012345678901234", "+0067360b70e4acfab561fe593ad3f7a"})
```
ahoffer: An advantage of making software components is that the test logic can focus on just the component. Components can be unit tested instead of having to test behavior through web services. It looks like most of what is exercised in this test is the plumbing-- the parts of Tomcat and Spring that handle HTTP requests and dispatch them to controllers. We generally assume the plumbing works and only test it in a few high-level tests. That leaves developers free to write more unit tests. The unit tests are simpler and faster to run as well as to write.

src/main/java/com/connexta/search/index/controllers/MultipartFileValidator.java
emmberk: What would be the value of injecting it in the bean? Would we ever mock it? Is it more efficient?

src/test/java/com/connexta/search/index/IndexTests.java
emmberk: I don't see what other "high-level tests" are verifying this if not this one. I also don't think we can make the assumption that "the plumbing works" because we had to build in a custom ExceptionHandler to handle this case, which wouldn't have been caught without a test like this.

PR #70, opened 2019-11-04
New attribute names and their test values

ahoffer:
[READY TO REVIEW]
I want to let the team get a look at what was chosen for attribute names and for test data. That way, we can make sure everyone is happy with them before early in the sprint. @mcalcote  and I coordinated on the names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,9 +128,16 @@ The following are valid query attributes:
 
 | Attribute | Description  |
 |---|---|
-| `id` | The unique ID that is associated with each dataset. |
-| `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the IRM. |
+| `contents` | A client can perform keyword queries with this attribute |
+| `country_code` | GENC:3:3-7 trigraph indicating a country |
+| `created` | The date the File was created |
+| `expiration` | The date past which the information in the Dataset should not be consumed |
+| `icid` | The Information Community Identification of the File |
+| `keyword` | The topic of the Dataset |
+| `media_type` | The media (MIME) type of the File |
+| `modified` | The date the File was last modified |
+| `resource_uri` | Location of the File |
+| `title` | Name given to the Dataset |
 
 ### Inspecting
 The service is deployed with (Springfox) **Swagger UI**.
--- a/src/main/java/com/connexta/search/index/Index.java
+++ b/src/main/java/com/connexta/search/index/Index.java
@@ -6,9 +6,9 @@
  */
 package com.connexta.search.index;
 
-import static com.connexta.search.common.configs.SolrConfiguration.CONTENTS_ATTRIBUTE_NAME;
-import static com.connexta.search.common.configs.SolrConfiguration.ID_ATTRIBUTE_NAME;
-import static com.connexta.search.common.configs.SolrConfiguration.MEDIA_TYPE_ATTRIBUTE_NAME;
+import static com.connexta.search.common.configs.SolrConfiguration.CONTENTS_ATTRIBUTE;
+import static com.connexta.search.common.configs.SolrConfiguration.ID_ATTRIBUTE;
+import static com.connexta.search.common.configs.SolrConfiguration.MEDIA_TYPE_ATTRIBUTE;
 
 import com.connexta.search.common.configs.SolrConfiguration;
 import lombok.AllArgsConstructor;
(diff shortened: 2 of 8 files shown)


jhunzik:
My only suggestion is that maybe we add a `_DATE` suffix to the date attributes:
`CREATED_ATTRIBUTE_NAME = "created"` -> `CREATED_DATE_ATTRIBUTE_NAME = "created_date"`

README.md
```diff
@@ -128,9 +128,17 @@ The following are valid query attributes:
 
 | Attribute | Description  |
 |---|---|
-| `id` | The unique ID that is associated with each dataset. |
-| `contents` | A client can perform keyword queries with this attribute. |
-| `media_type` | The media (MIME) type of the IRM. |
+| `id` | The unique ID that is associated with each Dataset |
```
peterhuffer: ❓ Should we also be documenting the types of these attributes?

src/main/java/com/connexta/search/index/Index.java
```diff
@@ -25,12 +25,12 @@
 @NoArgsConstructor
 public class Index {
```
peterhuffer: ❓ This class needs to be expanded to include all the new attributes, right?

src/test/java/com/connexta/search/common/SearchTestData.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class SearchTestData {
+
+  @Nullable
+  public static String get(String key) {
+    return map.get(key);
+  }
+
+  public static final Map<String, String> map =
+      ImmutableMap.<String, String>builder()
```
peterhuffer: ✏️ Java 9 added `Map.of()` for immutable maps.

src/test/java/com/connexta/search/query/GetFilterTests.java
```diff
@@ -39,4 +46,18 @@ public void testMalformedQuery() {
...
         MalformedQueryException.class, () -> QueryServiceImpl.getFilter("contents LIKE don't"));
   }
+
+  @Test
+  void testAcceptsAllValidQueryAttributes() {
+
+    // Assert that test data contains all valid terms.
+    Assert.assertThat(
+        "Sample data must include all query attributes",
+        unmodifiableSet(SearchTestData.map.keySet()),
```
peterhuffer: ✏️ `Set.of()`

src/test/java/com/connexta/search/query/GetFilterTests.java
```diff
@@ -39,4 +46,18 @@ public void testMalformedQuery() {
...
+        equalTo(QUERY_TERMS));
+
+    // Assert all terms are accepted
+    Filter filter = QueryServiceImpl.getFilter(SearchTestData.allAttributesQuery);
```
peterhuffer: I deleted this class in my PR, as well as removed the getFilter method. I think this test is more appropriate to be tested through the query method. We care more about the query method accepting these attributes.

src/main/java/com/connexta/search/index/Index.java
peterhuffer: My PR adds the solr schema which will need to be modified to allow accepting of these attributes. I also introduced solr itests to test the solr schema that will need to be modified.

README.md
ahoffer: That is a good question. Should we? My opinion, now that I think about it, is "no".

src/main/java/com/connexta/search/index/Index.java
ahoffer: Another great question. And one I hope to punt until the ticket in the next sprint that adds indexing for the new attributes.

README.md
peterhuffer: Since the CST are our public API, it needs to be documented somewhere. I don't know if a README is the best place for it.

ahoffer: I don't think searching on DatasetId is part of the CST

ahoffer: "content" is not part of the CST either. I left it because our test cases depend on it. Even manual testing depends on it.

src/test/java/com/connexta/search/common/SearchTestData.java
ahoffer: @peterhuffer and I looked at this and found a way to do it with Map.forEntries.

src/test/java/com/connexta/search/query/GetFilterTests.java
ahoffer: Slide over and give me your recommendation.

ahoffer: I'll move this test to go through the public method when @peterhuffer's code is merged and I rebase.

ahoffer:
@connexta/i2o-cdr After some team discussion we decided to include modifying the Solr schema (now that it exists) to include new query terms. That will be in a separate PR. Also, we agreed to add additional terms if there was time to do so later in the sprint.

PR #72, opened 2019-11-07
Add code coverage

peterhuffer:
- Adds jacoco
- Upgrades gradle wrapper
- Add gradle properties file

For some reason coverage includes itests. I'm not sure if its the naming schema we are using.

To hero:
- Full build
- Verify files in `build/reports/jacoco`, mainly index.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,8 +20,9 @@ buildscript {
 
 plugins {
     id "com.diffplug.gradle.spotless" version "3.25.0"
-    id "java"
     id "com.palantir.docker" version "0.22.1"
+    id "jacoco"
+    id "java"
     id "maven"
     id "org.owasp.dependencycheck" version "5.2.2"
     id "org.springframework.boot" version "${springBootVersion}"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -80,6 +81,30 @@ targetCompatibility = 1.11
...
     useJUnitPlatform()
+    finalizedBy(['jacocoTestReport', 'jacocoTestCoverageVerification'])
+}
+
+jacoco {
+    toolVersion = "0.8.5"
+    reportsDir = file("$buildDir/reports/jacoco")
+}
+
+jacocoTestReport {
```
emmberk: ✏️ Add a note to the readme about where the coverage report can be found, like we do with the owasp report.

build.gradle
```diff
@@ -80,6 +81,30 @@ targetCompatibility = 1.11
...
+        csv.enabled false
+        html.destination file("${buildDir}/reports/jacoco")
+    }
+}
+
+jacocoTestCoverageVerification {
+    violationRules {
+        rule {
+            limit {
+                minimum = 0.72
```
emmberk: ❓These questions shouldn't block this PR. I'm just curious. How does this work for submodules? Is all of the java code analyzed regardless of which module it's in and that all is compiled into one report? I know the solr submodule doesn't have any tests in it now, but if it did, would it be analyzed?

peterhuffer: I'm not exactly sure, but I'd assume it would work by just overriding these rules in the submodules gradle file by redefining this task for that sub module.

peterhuffer: So each module would have its own jacoco task, something like `:module1:jacocoTestCoverageVerification` and `:module2:jacocoTestCoverageVerification`. It's possible to tell the top level gradle file to run certain tasks on all sub modules

PR #75, opened 2019-11-08
Add Kubernetes deployment manifest

jhunzik:
[Ready for Review]
This PR gives us a Kubernetes deployment option. This will mainly be used for remote deployment, but you can deploy locally if you use Minikube as you local cluster.

Deployment is mostly automated using `./gradlew deployK8s`
In order to deploy remotely, you will need to create a `K8S_DOCKER_REGISTRY` environment variable and set `kubectl`'s context to the remote registry (see README)

This should be tested by deploying locally or remotely on Mac, Linux, and Windows.


.k8s/deploy-k8s.groovy
```diff
@@ -0,0 +1,67 @@
+//Groovy scripts throw lots of WARNING messages, but it is fine
+//https://issues.apache.org/jira/browse/GROOVY-8339
+
+
+def DOCKER_REG = System.getenv("DOCKER_REGISTRY")
```
ahoffer: Might want to tell people to set `DOCKER_REGISTRY` and `DOCKER_WRAPPER` in the README. If it isn't already in some other README. If you do, let people know to use the full path to the docker wrapper file.

.k8s/deploy-k8s.groovy
```diff
@@ -0,0 +1,67 @@
...
+    println ""
+    println "# # # # # # # # # # # # # #"
+    println " " + message
+    println ""
+}
+
+checkVars(DOCKER_W)
+
+
+run("Tagging Docker images with: " + DOCKER_REG)
```
ahoffer: This line does not run on my machine (linux). I get an error:
`Caught: java.io.IOException: Cannot run program "Tagging": error=2, No such file or directory`

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+To work with a K8s cluster you need to install Kubectl, a tool for controlling your K8s deployment.
+[Kubectl installation guide](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+
+### Before Deploying
+Kubernetes manifests do not support adding config files directly in the manifest via a file path. Instead we need to use
+a resource called ConfigMaps.
+
+ConfigMaps are how we pass configurations to the containers. These can be created through literals or files. The only
+issue is that these can"t be created in a manifest based off a file like we used to do with Docker Compose. Now we have
+to created them manually using `kubectl create configmap <config-map-name> --from-file=<path-to-file>`
```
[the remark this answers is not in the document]
jhunzik: Good point, I'll update that.

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+### Deploying
+The deployment manifest file contains everything to deploy our application to the Kubernetes cluster. To deploy use:
+`kubectl apply -f search-deployment.yml`
+
+To remove a deployment and all related pods and containers use:
+`kubectl delete -f search-deployment.yml`
+
+Note: The ConfigMap will not be deleted automatically because it was added to the cluster manually.
+#### Optional:
+Install MiniKube, a tool for running Kubernetes locally in a VM or the host
```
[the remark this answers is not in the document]
jhunzik: Yeah, to run locally we would need to use Minikube. Kubectl is the tool that is used to interact with a Kubernetes cluster (ie Minikube)

.k8s/deploy-k8s.groovy
jhunzik: We actually won't need the wrapper, at least for now. I'll update the readme to mention the env variable.

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
+# Kubernetes Deployment
+## Deployment
+### The Tool
+To work with a K8s cluster you need to install Kubectl, a tool for controlling your K8s deployment.
+[Kubectl installation guide](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+
+### Before Deploying
+Kubernetes manifests do not support adding config files directly in the manifest via a file path. Instead we need to use
```
peterhuffer: ✏️ I would avoid using words like `we` or `our` and opt for something like `ConfigMaps are instead used to add configs.` And below this `ConfigMaps are used to pass configurations to the containers.`

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+a resource called ConfigMaps.
+
+ConfigMaps are how we pass configurations to the containers. These can be created through literals or files. The only
+issue is that these can"t be created in a manifest based off a file like we used to do with Docker Compose. Now we have
```
peterhuffer: ✏️ `can"t` -> `can't`

.k8s/K8s-Deployment.md
peterhuffer: ✏️ `created` -> `create`

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+To remove a deployment and all related pods and containers use:
+`kubectl delete -f search-deployment.yml`
+
+Note: The ConfigMap will not be deleted automatically because it was added to the cluster manually.
+#### Optional:
+Install MiniKube, a tool for running Kubernetes locally in a VM or the host
+    machine. Deploys a lightweight single-node k8s cluster.
+Note: Run `eval $(minikube docker-env)` to enter Minikube"s docker context
+
+When starting up [Minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/), make sure your vm-driver is the one bundled with
```
peterhuffer: ✏️ Probably put this link to install minikube at the first mention of minikube above rather than here.

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+VirtualBox. Dnsmasq doesn"t play nice with the standalone hyperkit.
```
peterhuffer: ✏️ `doesn"t` -> `doesn't`

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,112 @@
...
+metadata:
+    name: solr
+    labels:
+        app: solr
+spec:
+    type: ClusterIP
+    selector:
+        app: solr
+    ports:
+      - port: 8983
```
peterhuffer: ❓ Did we expose 9983 before for a reason?

.k8s/K8s-Deployment.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Command Cheat Sheet
+- `kubectl cluster-info`: View Cluster info
+- `kubectl get` - List resources
+    - `kubectl get nodes`: Lists available Nodes in the cluster
+    - `kubectl get pods`: Lists Pods in your cluster
+    - `kubectl get deployments`: List deployments in your cluster
+- `kubectl describe` - Show detailed information about a resource
+- `kubectl logs` - Print the logs from a container in a pod
+- `kubectl exec` - Execute a command on a container in a pod
```
figliold: pod"s should be pod's and there"s should be there's

.k8s/deploy-k8s.groovy
```diff
@@ -0,0 +1,67 @@
...
+}
+
+//The name of the preface for the docker images
+def DOCKER_IMAGE_NAME = "cnxta/ion-search"
+def DOCKER_SOLR_IMAGE = "cnxta/search-solr"
+
+//The path to the Docker Wrapper
+//IMPORTANT!
+//>>> The base docker script has to be in the same location as this file
+
```
figliold: ❓ If we don't need the docker wrapper, should we remove references to it in these scripts?

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,112 @@
...
+    selector:
+        matchLabels:
+            app: search
+    template:
+        metadata:
+            labels:
+                app: search
+        spec:
+            containers:
+              - name: search
```
figliold: ❓ Is "docker.io" for dockerhub images?  If so, will these images ever be on dockerhub?

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,112 @@
...
+#     selector:
+#         matchLabels:
+#             app: solr
+#     template:
+#         metadata:
+#             labels:
+#                 app: solr
+#         spec:
+#             containers:
+#               - name: solr
```
figliold: Same comment about dockerhub. See above.

Dockerfile
```diff
@@ -4,6 +4,7 @@ LABEL com.connexta.application.name=search
 ARG JAR_FILE
 COPY ${JAR_FILE} /search
 ENTRYPOINT ["/search"]
+RUN chmod 755 /search
 EXPOSE 8080 10042 10052
```
figliold: Not that removing it should be part of this ticket, but are we using any of the JMX stuff? https://github.com/connexta/ion-search/pull/75/files#diff-3254677a7917c6c01f55212f86c57fbfR12-R18

.k8s/search-deployment.yml
jhunzik: That's the port we chose to expose on the container. Solr was still running on 8983 internally.

.k8s/K8s-Deployment.md
jhunzik: I need to update the formatter to exclude `.md` files

.k8s/deploy-k8s.groovy
jhunzik: Yeah, I'll remove that in the next commit

.k8s/search-deployment.yml
jhunzik: No, we have a registry that we'll push them to just like we did for our other deployment

Dockerfile
jhunzik: I think so, for live debugging.

.k8s/deploy-k8s.groovy
jhunzik: If made some changes since then, I think you should give it another try

.k8s/search-deployment.yml
jhunzik: I changed this to tag it to the correct registry.

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,143 @@
...
+              cpu: "1"
+              memory: "4Gi"
+          ports:
+            - containerPort: 8983
+              protocol: TCP
+          volumeMounts:
+            - name: solr-pvc
+              mountPath: /var/solr/data
+      volumes:
+        - name: solr-pvc
```
aperullo: Nit-picking, but this is the persistent volume itself, not the pvc. This might be better as `solr-pv`.

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,143 @@
...
+          persistentVolumeClaim:
+            claimName: pvc
```
aperullo: Then this should be `solr-pvc`

.k8s/search-deployment.yml
```diff
@@ -0,0 +1,143 @@
...
+  selector:
+    app: solr
+  ports:
+    - port: 8983
+# Creates a Persistent Volume Claim to persist Solr core data across deployments
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pvc
```
aperullo: and this should also become `solr-pvc`.

.k8s/search-deployment.yml
jhunzik: Good catch!

PR #81, opened 2019-11-20
Index additional attributes

ahoffer:
[Ready to Review]
This PR is only part of a feature. Follow-on PRs will add more attributes.

- Adds dates as well as strings to Index and solr schema
- Renamed irmUriString to irmUri
- Switches construction of Index objects to the builder pattern using Lombok's `@Builder`
- Adds unit test for query-able attributes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,6 @@ The following are valid query attributes:
 | `id` | The unique ID that is associated with each Dataset |
 | `keyword` | The topic of the Dataset |
 | `modified` | The date the File was last modified |
-| `irm_url` | Location of the IRM document |
 | `title` | Name given to the Dataset |
 
 ### Inspecting
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ The following are valid query attributes:
 | `id` | The unique ID that is associated with each Dataset |
 | `keyword` | The topic of the Dataset |
 | `modified` | The date the File was last modified |
-| `resource_uri` | Location of the File |
+| `irm_url` | Location of the IRM document |
 | `title` | Name given to the Dataset |
 
 ### Inspecting
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ jacocoTestCoverageVerification {
     violationRules {
         rule {
             limit {
-                minimum = 0.80
+                minimum = 0.77
             }
         }
     }
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ jacocoTestCoverageVerification {
     violationRules {
         rule {
             limit {
-                minimum = 0.77
+                minimum = 0.80
             }
         }
     }
--- a/solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
+++ b/solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
@@ -120,11 +120,11 @@
   <field name="contents" type="text_general" multiValued="false" required="true" />
   <field name="country_code" type="string" />
   <field name="created" type="pdate" sortMissingLast="true" />
-  <field name="file_url" type="string" required="false" /> <!--TODO: Required? -->
+  <field name="file_url" type="string" indexed="false" required="false" /> <!--TODO: Required? -->
   <field name="icid" type="string" />
-  <field name="irm_url" type="string" required="true" />
+  <field name="irm_url" type="string" indexed="false" required="true" />
   <field name="keyword" type="string" />
-  <field name="metacard_url" type="string" required="false" /> <!--TODO: Required?-->
+  <field name="metacard_url" type="string" indexed="false" required="false" /> <!--TODO: Required?-->
   <field name="modified" type="pdate"/>
   <field name="date_range" type="pdate" />
   <field name="end" type="pdate" />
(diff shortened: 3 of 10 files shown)


solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
```diff
@@ -116,19 +116,25 @@
...
+  <!-- CST fields -->
+  <field name="contents" type="text_general" indexed="true" stored="true" required="true"
+    multiValued="false"/>
+  <field name="country_code" type="string" indexed="true" stored="true" required="false"
+    multiValued="false"/>
+  <field name="created" type="pdate" sortMissingLast="true"
+    indexed="true" stored="true" required="false" multiValued="false"/>
+  <field name="icid" type="string" indexed="true" stored="true" required="false"
+    multiValued="false"/>
+  <field name="irm_uri" type="string" indexed="true" stored="true" required="true"
```
jhunzik: ✏️ Should we be using `url` here instead? https://github.com/connexta/ion-index-api/pull/35#discussion_r347642762

ahoffer: A fair question. I called it that because that was what is was called in SolrConfiguration. I think URL makes more sense. Here is the relevant snippet of DDF documentation (not sure that makes any difference to us, but here it is anyway)
image

ahoffer: If even one person says "change it to say `resource_url`, I'd do it.

emmberk: This is not related to DDF. We should use URL from now on per Eric.

README.md
```diff
@@ -128,7 +128,7 @@ The following are valid query attributes:
 | `id` | The unique ID that is associated with each Dataset |
 | `keyword` | The topic of the Dataset |
 | `modified` | The date the File was last modified |
-| `resource_uri` | Location of the File |
+| `irm_url` | Location of the IRM document |
```
ahoffer: :question:  Should URLs go in this list?

build.gradle
```diff
@@ -107,7 +107,7 @@ jacocoTestCoverageVerification {
     violationRules {
         rule {
             limit {
-                minimum = 0.80
+                minimum = 0.77
```
ahoffer: I suspect coverage when down because of the Lombok @Builder annotation. It generated a bunch of code and JaCoCo is not ignoring it. I read a bout a `lombok.config` file that was supposed to help JaCoCo ingore Lombok generated code, but it did not affect the coverage numbers. 

The coverage is low in the class that got the @Builder annotation:
image

solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
```diff
@@ -116,19 +116,20 @@
...
-           multiValued="false" />
-
-    <!-- If you don't use child/nested documents, then you should remove the next two fields:  -->
-    <!-- for nested documents (minimal; points to root document) -->
-    <field name="_root_" type="string" indexed="true" stored="false" docValues="false" />
-    <!-- for nested documents (relationship tracking) -->
-    <field name="_nest_path_" type="_nest_path_" /><fieldType name="_nest_path_" class="solr.NestPathField" />
-
-    <field name="_text_" type="text_general" indexed="true" stored="false" multiValued="true"/>
+  <!-- CST fields -->
```
ahoffer: Added some fields. Use default field properties to simplify the declarations.
See https://lucene.apache.org/solr/guide/8_1/field-type-definitions-and-properties.html#field-default-properties

src/test/java/com/connexta/search/common/SearchManagerImplTest.java
```diff
@@ -202,16 +222,41 @@ void testQueryInvalidEcql() {
...
+    queryPairs.put("modified", "2019-11-14");
+    queryPairs.put("title", "A Title");
+
+    assertThat(
+        "Expected attributes do not match actual attributes. These collections need to be the same.",
+        List.of(queryPairs.keySet()),
+        containsInAnyOrder(SolrConfiguration.QUERY_TERMS));
+
+    List<String> queryPreds = new ArrayList<>();
+    queryPairs.forEach((k, v) -> queryPreds.add(String.format(template, k, v)));
```
mcalcote: ❓ What is the purpose of `queryPreds`?

ahoffer: Hmmm. Looks like it shouldn't be there anymore now that the entrySet().stream() is working.

README.md
figliold: ❓ Should the irm url be queryable?  Seems like we would store it but we wouldn't query allow queries by it.

jhunzik: I'm not sure if it makes sense to query on a URL.

build.gradle
jhunzik: A quick google suggests that we can ignore lombok generated code in jacoco: https://www.rainerhahnekamp.com/en/ignoring-lombok-code-in-jacoco/

README.md
ahoffer: The vox populi has spoken! I'll pull the IRM URL out of the queryable attributes list.
I'll update the Solr schema accordingly.

build.gradle
ahoffer: I did try that and either:
1. I did it wrong
2. It doesn't work for some reason

PR #86, opened 2019-11-25
Content extractor, additional attributes

ahoffer:
This PR is ready for review.

* This PR adds some new indexable attributes to the Index class and updates the Solr schema to accommodate them.
* It adds two attributes that which are persistent, but not searchable: File URL and Metacard URL
* It changes the test code to use Lombok's @Builder instead of @AllArgsConstructor to create instances of Index
* @figliold It brings in the BodyContentExtractor. When this PR is merged, PRs 80 and 82 can be closed.
* Adds test for indexing and querying extracted content.
* @figliold and I got the StoreITests by updating them to account for both the IRM and File.


build.gradle
```diff
@@ -104,7 +120,8 @@ jacocoTestCoverageVerification {
     violationRules {
         rule {
             limit {
-                minimum = 0.77
+                minimum = 0.68
+                // Dan and Aaron took at quick look at coverage. Mostly missing exception blocks. There is a ticket to redo exception handling which could solve the issue.
```
jhunzik: Coverage was at 86% when we finished yesterday. We should update this. 🎉

src/main/java/com/connexta/search/common/Index.java
```diff
@@ -20,15 +20,19 @@
...
 import org.springframework.data.annotation.Id;
 import org.springframework.data.solr.core.mapping.Indexed;
 import org.springframework.data.solr.core.mapping.SolrDocument;
 
 @SolrDocument(collection = SOLR_COLLECTION)
 @Getter
+@Setter
 @EqualsAndHashCode
 @Builder(toBuilder = true)
+@Generated
```
jhunzik: A note to other reviewers, this annotation flags the class as generated and Jacoco will exclude it from it's coverage analysis.

src/test/java/com/connexta/search/SearchITest.java
```diff
@@ -557,6 +325,23 @@ public void describeTo(final Description description) {
     };
   }
 
+  @NotNull
```
jhunzik: ✏️ We should move this to the bottom with the other helper mehtods

ahoffer:
OWASP passes, but Snyk fails. What course of action to we take when that happens?

ahoffer:
@connexta/i2o-cdr I pushed a commit that makes the URL fields in Solr unsearchable. That means the user can't query them, but it also means we can't query for them either. Is there ever a trouble-shotting or debugging scenario when we might want to search for a record by one of its related URLs?

build.gradle
```diff
@@ -46,6 +46,22 @@ dependencies {
     implementation "io.springfox:springfox-swagger2:${springfoxSwaggerVersion}"
     implementation "io.springfox:springfox-swagger-ui:${springfoxSwaggerVersion}"
     implementation "javax.inject:javax.inject:1"
+    implementation("org.apache.tika:tika-parsers:1.22") {
```
emmberk: @ahoffer Starting a thread to respond to https://github.com/connexta/ion-search/pull/86#issuecomment-558728222:

We've been ignoring snyk findings so far across all of our repos. They are not required checks for PRs.

solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
```diff
@@ -116,15 +116,15 @@
...
   <field name="country_code" type="string" />
   <field name="created" type="pdate" sortMissingLast="true" />
-  <field name="file_url" type="string" indexed="false" required="false" /> <!--TODO: Required? -->
+  <field name="file_url" type="string" required="false" indexed="false" /> <!--TODO: Required? -->
   <field name="icid" type="string" />
-  <field name="irm_url" type="string" indexed="false" required="true" />
+  <field name="irm_url" type="string" required="true" indexed="false" />
   <field name="keyword" type="string" />
-  <field name="metacard_url" type="string" indexed="false" required="false" /> <!--TODO: Required?-->
+  <field name="metacard_url" type="string" required="false" indexed="false" /> <!--TODO: Required?-->
```
emmberk: @ahoffer Starting a thread to respond to https://github.com/connexta/ion-search/pull/86#issuecomment-558729044

I don't think we'd ever want to search by the URLs.

src/main/java/com/connexta/search/index/BodyContentExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+    tika = new Tika();
+    tika.setMaxStringLength(NO_LIMIT_MAX_CONTENT_LENGTH);
+  }
+
+  public BodyContentExtractor(int maxContentLength) {
+    this.tika = new Tika();
+    tika.setMaxStringLength(maxContentLength);
+  }
+
+  public String extractText(InputStream inputStream) throws IOException, TikaException {
```
emmberk: ❓ Is tika thread-safe?

src/main/java/com/connexta/search/index/BodyContentExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class BodyContentExtractor {
+
+  private static final int NO_LIMIT_MAX_CONTENT_LENGTH = -1;
+
+  private final Tika tika;
+
+  public BodyContentExtractor() {
+    tika = new Tika();
+    tika.setMaxStringLength(NO_LIMIT_MAX_CONTENT_LENGTH);
```
emmberk: ✏️ Should we limit this be default?

src/main/java/com/connexta/search/index/BodyContentExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+  }
```
emmberk: ✏️ 
```java
public BodyContentExtractor() {
  this(NO_LIMIT_MAX_CONTENT_LENGTH);
}
```

src/test/java/com/connexta/search/index/BodyContentExtractorTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private static final String SAMPLE_JPG = "file_example_JPG_100kB.jpg";
+
+  private static final String SAMPLE_TXT = "sample1.txt";
+
+  @Test
+  @DisplayName("Text Extraction from PDF")
+  public void testExtractTextPdf() throws Exception {
+    final String EXPECTED_WORD = "demonstration";
+    final BodyContentExtractor bodyContentExtractor = new BodyContentExtractor();
```
emmberk: ✏️ If testMaxContentLength is removed (see other comment), bodyContentExtractor can be made a class field instead of constructing it in every test method.

src/test/java/com/connexta/search/index/BodyContentExtractorTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private static final String SAMPLE_XLSX = "file_example_XLSX_10.xlsx";
+
+  private static final String SAMPLE_JPG = "file_example_JPG_100kB.jpg";
+
+  private static final String SAMPLE_TXT = "sample1.txt";
+
+  @Test
+  @DisplayName("Text Extraction from PDF")
+  public void testExtractTextPdf() throws Exception {
```
emmberk: ✏️ Can all of these extract tests be parameterized?

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
   @Test
-  void testNullResource(
-      @Mock final WebClient mockWebClient,
-      @Mock final RequestHeadersUriSpec mockRequestHeadersUriSpec,
-      @Mock final RequestHeadersSpec mockRequestHeadersSpec,
-      @Mock final ResponseSpec mockResponseSpec,
-      @Mock final ResponseSpec mockResponseSpec2,
-      @Mock final Mono<Resource> mockMono)
-      throws Exception {
+  void testDatasetAlreadyExists(@Mock final InputStream mockInputStream) {
```
emmberk: ✏️ mockInputStream isn't be used

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-  void testCouldNotOpenInputStream(
-      @Mock final WebClient mockWebClient,
-      @Mock final RequestHeadersUriSpec mockRequestHeadersUriSpec,
-      @Mock final RequestHeadersSpec mockRequestHeadersSpec,
-      @Mock final ResponseSpec mockResponseSpec,
-      @Mock final ResponseSpec mockResponseSpec2,
-      @Mock final Mono<Resource> mockMono,
-      @Mock final Resource mockResource)
-      throws Exception {
+  void testExceptionWhenCheckingIfDatasetExists(@Mock final InputStream mockInputStream) {
```
emmberk: ✏️ mockInputStream isn't be used

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-                      .index(
-                          datasetId,
-                          storeMockWebServer
-                              .url(String.format("/dataset/%s/irm", datasetId))
-                              .uri()));
-    }
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
+  @AfterEach
+  void afterEach() {
+    reset(mockBodyContentExtractor, mockIndexRepository, mockGetInputStream);
```
emmberk: ✏️ I think it's simpler and more readable to use mockito's annotation for creating mocks:

```java
@Mock private BodyContentExtractor mockBodyContentExtractor;
@Mock private IndexRepository mockIndexRepository;
@Mock private IonResourceLoader mockIonResourceLoader;

private IndexService indexService;

@BeforeEach
void beforeEach() {
  indexService = new IndexServiceImpl(mockIndexRepository, mockGetInputStream, mockBodyContentExtractor);
}
```

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+  static final String FILE_URL = "http://file";
+  static final String IRM_URL = "http://irm";
 
-  @AfterEach
-  public void afterEach() {
-    verifyNoMoreInteractions(mockSearchManager);
-  }
+  final BodyContentExtractor mockBodyContentExtractor = mock(BodyContentExtractor.class);
+  final IndexRepository mockIndexRepository = mock(IndexRepository.class);
+  final IonResourceLoader mockGetInputStream = mock(IonResourceLoader.class);
```
emmberk: ✏️ `mockIonResourceLoader`

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
 
-  @ParameterizedTest
-  @ValueSource(ints = {400, 401, 403, 500, 501})
-  void testIrmUriReturnsNotOkStatusCode(final int code) throws Exception {
-    final String datasetId = "00067360b70e4acfab561fe593ad3f7a";
+  final IndexService indexService =
+      new IndexServiceImpl(mockIndexRepository, mockGetInputStream, mockBodyContentExtractor);
```
emmberk: ✏️ All of these fields can be private

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-    when(mockRequestHeadersUriSpec.uri(irmUri)).thenReturn(mockRequestHeadersSpec);
-    when(mockRequestHeadersSpec.retrieve()).thenReturn(mockResponseSpec);
-    when(mockResponseSpec.onStatus(any(Predicate.class), any(Function.class)))
-        .thenReturn(mockResponseSpec2);
-    when(mockResponseSpec2.bodyToMono(Resource.class)).thenReturn(mockMono);
-
-    final RuntimeException runtimeException = new RuntimeException();
-    when(mockMono.block()).thenThrow(runtimeException);
+    indexService.index(Index.builder().id(DATASET_ID).fileUrl(FILE_URL).irmUrl(IRM_URL).build());
+  }
```
emmberk: ❗️This test should verify 

- the correct URL was used in the loader
- the inputstream from the loader is the one used by the contentextractor
- the correct index was saved

```java
@Test
void testIndex(@Mock final InputStream mockInputStream) throws IOException, TikaException {
  doReturn(mockInputStream)
      .when(mockIonResourceLoader)
      .get(FILE_URL);
  doReturn("fileContents").when(mockBodyContentExtractor).extractText(mockInputStream);
  when(mockIndexRepository.existsById(datasetId)).thenReturn(false);

  indexService.index(Index.builder().id(DATASET_ID).fileUrl(FILE_URL).irmUrl(IRM_URL).build());

  verify(mockIndexRepository).save(TODO);
}
```

src/main/java/com/connexta/search/common/Index.java
```diff
@@ -20,15 +20,19 @@
...
 public class Index {
```
emmberk: ✏️ Add `@ToString`

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+  static final String DATASET_ID = "00067360b70e4acfab561fe593ad3f7a";
+  static final String FILE_URL = "http://file";
+  static final String IRM_URL = "http://irm";
 
-  @AfterEach
-  public void afterEach() {
-    verifyNoMoreInteractions(mockSearchManager);
-  }
+  final BodyContentExtractor mockBodyContentExtractor = mock(BodyContentExtractor.class);
+  final IndexRepository mockIndexRepository = mock(IndexRepository.class);
```
emmberk: ❗️ Verify no more interactions with the mockIndexRepository. We should protect against saving random indexes.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-                ((IndexService) new IndexServiceImpl(mockSearchManager, mockWebClient))
-                    .index(datasetId, irmUri));
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
-    assertThat(thrown.getCause(), is(runtimeException));
+  @Test
+  void testBadInputStreamWhenProcessingIrm() throws IOException {
+    doThrow(IOException.class).when(mockGetInputStream).getAsString(anyString());
+
+    assertThrows(
+        SearchException.class,
```
emmberk: ❗️Verify that the status code and cause of the thrown exception

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-    final SearchException thrown =
-        assertThrows(
-            SearchException.class,
-            () ->
-                ((IndexService) new IndexServiceImpl(mockSearchManager, mockWebClient))
-                    .index(datasetId, irmUri));
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
+    // expect
+    assertThrows(
+        SearchException.class, () -> indexService.index(Index.builder().id(datasetId).build()));
```
emmberk: ❗️Verify that the status code and cause of the thrown exception

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+    // expect
     final SearchException thrown =
         assertThrows(
-            SearchException.class,
-            () ->
-                ((IndexService) new IndexServiceImpl(mockSearchManager, mockWebClient))
-                    .index(datasetId, irmUri));
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
-    assertThat(thrown.getCause(), is(ioException));
+            SearchException.class, () -> indexService.index(Index.builder().id(datasetId).build()));
```
emmberk: ❗️Verify that the status code of the thrown exception

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-                  ((IndexService) new IndexServiceImpl(mockSearchManager, WebClient.create()))
-                      .index(datasetId, irmUri));
-    }
-
-    assertThat(
-        "thrown exception is the exact same exception thrown by the SearchManager",
-        thrown,
-        is(runtimeException));
+    final URI irmUri = new URI(String.format("http://store:9041/dataset/%s/irm", datasetId));
+    final URI fileUri = new URI(String.format("http://store:9041/dataset/%s/", datasetId));
```
emmberk: ✏️ No need to construct URIs and then covert to strings. These can also be inlined.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+        .when(mockIndexRepository)
+        .save(
+            argThat(
+                anIndex -> {
+                  return anIndex.equals(index);
+                }));
+
+    // expect
+    final SearchException thrown =
+        assertThrows(SearchException.class, () -> indexService.index(index));
```
emmberk: ❗️Verify that the status code of the thrown exception

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+
+    // and stub CrudRepository#save
+    final RuntimeException runtimeException = new RuntimeException();
+    doThrow(runtimeException)
+        .when(mockIndexRepository)
+        .save(
+            argThat(
+                anIndex -> {
+                  return anIndex.equals(index);
+                }));
```
emmberk: ✏️ 
```java
.save(eq(index));
```

You also don't need to explicitly check the argument here because that's already tested in `testIndex`.
```java
.save(any());
```
is good enough.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
-
-    final ArgumentCaptor<InputStream> inputStreamCaptor =
-        ArgumentCaptor.forClass(InputStream.class);
-    verify(mockSearchManager).index(eq(datasetId), eq(irmUri), inputStreamCaptor.capture());
-    assertThat(IOUtils.toString(inputStreamCaptor.getValue(), StandardCharsets.UTF_8), is(body));
+  void testUnableToReadIrmBody() throws IOException {
+    final IOException ioException = new IOException();
+    doThrow(ioException).when(mockGetInputStream).getAsString(anyString());
+    assertThrows(
+        SearchException.class,
```
emmberk: ❗️Verify that the status code and cause of the thrown exception

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,175 @@
...
+
+  @ParameterizedTest
+  @MethodSource("exceptionsWhileExtractingText")
+  void testExceptionsWhileExtractingText(final Throwable exception)
+      throws IOException, TikaException {
+    final InputStream mockInputStream = mock(InputStream.class);
+    doReturn(mockInputStream).when(mockGetInputStream).get(FILE_URL);
+    doThrow(exception).when(mockBodyContentExtractor).extractText(any(InputStream.class));
+    assertThrows(
+        SearchException.class,
```
emmberk: ❗️Verify that the status code and cause of the thrown exception

owasp-suppressions.xml
```diff
@@ -1,3 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
+    <suppress>
```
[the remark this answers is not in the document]
ahoffer: Dan wrote these suppression. I copy what he wrote on Slack into the file.

src/main/java/com/connexta/search/common/Index.java
```diff
@@ -20,15 +20,19 @@
...
+import lombok.Generated;
 import lombok.Getter;
+import lombok.Setter;
 import org.springframework.data.annotation.Id;
 import org.springframework.data.solr.core.mapping.Indexed;
 import org.springframework.data.solr.core.mapping.SolrDocument;
 
 @SolrDocument(collection = SOLR_COLLECTION)
 @Getter
+@Setter
```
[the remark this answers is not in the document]
ahoffer: I like that.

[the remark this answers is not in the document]
ahoffer: I removed `toBuilder` and #testUpdate. You probably already know that `toBuilder` preserves immutability. `toBuilder`  makes a copy of the object and only changes values on the copy. The original object is not changed.

owasp-suppressions.xml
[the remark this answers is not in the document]
emmberk: I think these should be addressed before merging.

src/main/java/com/connexta/search/common/Index.java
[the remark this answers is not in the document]
emmberk: I didn't mean modify using the builder. I meant modify using the setters.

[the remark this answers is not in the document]
emmberk: I think the moving object along the conveyer belt should be the builder, not the Index object itself.

src/main/java/com/connexta/search/common/Index.java
jhunzik: Yeah, I checked the other annotations and they don't include `@Generated`

solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
peterhuffer: ❓ Does stored default to true?

src/main/java/com/connexta/search/index/IndexServiceImpl.java
```diff
@@ -7,70 +7,81 @@
...
+  @NotNull private final IndexRepository indexRepository;
+  @NotNull private final IonResourceLoader ionResourceLoader;
+  @NotNull private final BodyContentExtractor bodyContentExtractor;
 
   @Override
-  public void index(
-      @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId,
-      @NotNull final URI irmUri) {
-    final Resource resource;
+  public void index(Index index) {
```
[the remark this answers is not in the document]
emmberk: This list of parameter will not grow. It is not the same as everything that gets indexed. I still think that the controller shouldn't be aware at all about how things are indexed. If it should pass an object instead of a few params, I don't think that object should be the Index object or even an IndexBuilder. The IndexService should decide what gets indexed, not the Controller.

src/main/java/com/connexta/search/common/Index.java
emmberk: Oh, I didn't see the "lombok doesn't actually add this annotation" note 😆

src/main/java/com/connexta/search/common/Index.java
```diff
@@ -20,15 +20,19 @@
...
 @EqualsAndHashCode
 @Builder(toBuilder = true)
+@Generated
 public class Index {
 
   @Indexed(name = CONTENTS_ATTRIBUTE)
```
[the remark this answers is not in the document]
emmberk: You're right. `id`, `iurmUrl`, and `metacardUrl` may not be null. We decided to only mark Nullable stuff, so removing `@Nullable` from those three should be good. There should probably be tests to make sure that `iurmUrl` and `metacardUrl` are required as well.

src/main/java/com/connexta/search/index/IndexService.java
```diff
@@ -6,14 +6,10 @@
...
 import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Pattern;
-import javax.validation.constraints.Size;
 
 public interface IndexService {
 
-  void index(
-      @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId,
-      @NotNull final URI irmUri);
+  void index(@NotNull Index index);
```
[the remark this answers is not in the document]
emmberk: you added javadoc, but didn't document the exception.

src/main/java/com/connexta/search/index/IonResourceLoader.java
```diff
@@ -0,0 +1,28 @@
...
+public class IonResourceLoader {
+  private final ResourceLoader resourceLoader;
+
+  // TODO: Not intended to be permanent API
+  public String getAsString(String resourceUrl) throws IOException {
+    return IOUtils.toString(get(resourceUrl), StandardCharsets.UTF_8);
+  }
+
+  public InputStream get(String resourceUrl) throws IOException {
+    return resourceLoader.getResource(resourceUrl).getInputStream();
```
peterhuffer: ✏️ I'm not sure this will work for our services. This will be a URL like `http://blah:8080/dataset/abc123/irm`. That endpoint will require an accept version in the request headers.

src/main/java/com/connexta/search/index/BodyContentExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public BodyContentExtractor(int maxContentLength) {
```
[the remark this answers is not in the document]
emmberk: ❗️ I do see that one test, but it's not _fully_ tested. I would expected tests for negative, zero, and very large values.

src/main/java/com/connexta/search/index/IonResourceLoader.java
peterhuffer: Actually, since we do not define the retrieve endpoint in the API we won't need the accept version. But eventually this index service will need to be aware of the store version it is communicating with.

src/main/java/com/connexta/search/index/BodyContentExtractor.java
emmberk: ❗️The configuration class still uses the no-arg constructor which sets the limit to -1. I think this should be addressed before the PR is merged.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,163 @@
...
 
-  @Mock private SearchManager mockSearchManager;
+  static final String DATASET_ID = "00067360b70e4acfab561fe593ad3f7a";
+  static final String FILE_URL = "http://file";
+  static final String IRM_URL = "http://irm";
 
-  @AfterEach
-  public void afterEach() {
-    verifyNoMoreInteractions(mockSearchManager);
+  final ContentExtractorImpl mockContentExtractor = mock(ContentExtractorImpl.class);
```
emmberk: ✏️ The interface should be used here
```java
final ContentExtractor mockContentExtractor = mock(ContentExtractorImpl.class);
```

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,163 @@
...
-                              .url(String.format("/dataset/%s/irm", datasetId))
-                              .uri()));
-    }
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
+  @Test
+  void testIndex() throws IOException, TikaException {
+    doReturn(IOUtils.toInputStream("asdf", StandardCharsets.UTF_8))
+        .when(mockIonResourceLoader)
+        .get(anyString());
+    doReturn("fileContents").when(mockContentExtractor).extractText(any(InputStream.class));
```
emmberk: ❗️This should expect the same InputStream from the IonResourceLoader.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,163 @@
...
-                          storeMockWebServer
-                              .url(String.format("/dataset/%s/irm", datasetId))
-                              .uri()));
-    }
-    assertThat(thrown.getStatus(), is(HttpStatus.BAD_REQUEST));
+  @Test
+  void testIndex() throws IOException, TikaException {
+    doReturn(IOUtils.toInputStream("asdf", StandardCharsets.UTF_8))
+        .when(mockIonResourceLoader)
+        .get(anyString());
```
emmberk: ❗️ This should expect the correct URL

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
```diff
@@ -6,239 +6,163 @@
...
+    doReturn("fileContents").when(mockContentExtractor).extractText(any(InputStream.class));
+    indexService.index(testIndex);
+    verify(mockIndexRepository).save(testIndex);
```
emmberk: ❗️This should verify that the final index has the values that you expect, not just that is the same object that was originally passed in. This test should verify that the contents is populated with the expected value in the index that is saved.

src/test/java/com/connexta/search/index/IndexServiceImplTest.java
emmberk: ❗️This is still not complete.

src/main/java/com/connexta/search/common/Index.java
```diff
@@ -18,44 +18,66 @@
...
 import org.springframework.data.solr.core.mapping.SolrDocument;
 
 @SolrDocument(collection = SOLR_COLLECTION)
-@Getter
+@Builder
 @EqualsAndHashCode
-@Builder(toBuilder = true)
+@Generated
+@Getter
+@Setter
```
peterhuffer: ❓ Do we need the setter if we have the builder?

src/main/java/com/connexta/search/index/IonResourceLoader.java
emmberk: We currently don't use accept-versions for our retrieve endpoints.

src/main/java/com/connexta/search/common/Index.java
emmberk: I agree. I think Index should be immutable. See https://github.com/connexta/ion-search/pull/86#discussion_r350883675, https://github.com/connexta/ion-search/pull/86#discussion_r350886871, and https://github.com/connexta/ion-search/pull/86#discussion_r350901744.

src/main/java/com/connexta/search/index/IonResourceLoader.java
emmberk: @ahoffer, have you tested the ResourceLoader with out retrieve endpoints?

emmberk: Why will search need to be aware of the version in the future?

src/main/java/com/connexta/search/index/IndexServiceImpl.java
```diff
@@ -7,70 +7,81 @@
...
-import org.springframework.web.reactive.function.client.WebClient;
-import reactor.core.publisher.Mono;
 
 @Slf4j
 @AllArgsConstructor
 public class IndexServiceImpl implements IndexService {
 
-  @NotNull private final SearchManager searchManager;
-  @NotNull private final WebClient webClient;
+  @NotNull private final IndexRepository indexRepository;
```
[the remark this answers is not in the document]
emmberk: Ok. Just talked with @peterhuffer. Eric said that we should have a unique StorageAdaptor for each component. In our case, we have 2 components, index and query, each with their own StorageAdaptor, IndexRepository and StorageManager, respectively. So, I now think you did is good :+1: We can improve the naming and packages in another PR.

solr/src/main/resources/solr/conf/search-solr-managed-schema.xml
ahoffer: Yes. Here is the reference I'm using. https://lucene.apache.org/solr/guide/8_2/field-type-definitions-and-properties.html#field-default-properties

src/main/java/com/connexta/search/index/IonResourceLoader.java
figliold: Will there be any problems with the resource loader when we switch to https?

figliold: Do we get the resource from the URL every time we need an input stream?

